PR #1, opened 2016-11-14
Review

srush:
tmp


s2sa/encoder.lua
```diff
@@ -0,0 +1,64 @@
...
+
+  -- preallocate output gradients
+  self.grad_out_proto = {}
+  for _ = 1, args.num_layers do
+    table.insert(self.grad_out_proto, torch.zeros(args.max_batch_size, args.rnn_size))
+    table.insert(self.grad_out_proto, torch.zeros(args.max_batch_size, args.rnn_size))
+  end
+end
+
+function Encoder:forward(batch)
```
guillaumekln: I think we could but what would be the benefits in our case? Here `forward` is mostly a convenience method to factorize the input preparation during training, evaluation and translation.

s2sa/beam/main.lua
```diff
@@ -0,0 +1,810 @@
...
+          :view(1, batch_size, model_opt.rnn_size)
+          :expand(opt.beam, batch_size, model_opt.rnn_size)
+          :contiguous()
+          :view(opt.beam*batch_size, model_opt.rnn_size))
+    end
+  end
+end
+
+local function generate_beam(source, source_features, gold)
+  State.source_length = beam_utils.get_max_length(source)
```
guillaumekln: No, I got rid of it in a recent commit.

s2sa/generator.lua
```diff
@@ -0,0 +1,60 @@
...
+
+function Generator:training()
+  self.network:training()
+end
+
+function Generator:evaluate()
+  self.network:evaluate()
+end
+
+function Generator:convert(f)
```
guillaumekln: It is a function that convert all the attributes using the given function. See the base class `model.lua`.

srush: (add as a comment)

Martijnde:
Final review :)

guillaumekln:
Many of these comments were addressed in the past days either directly or indirectly with design changes.

Also, I would indeed like to see the following ideas happen in the future:

* Encoder/BiEncoder/Decoder as reusable `nn.Module` (i.e. override `updateOutput` instead of `forward` and `updateGradInput` and `accGradParameters` instead of `backward`) while still maximizing code factorization.
* Use the [optim](https://github.com/torch/optim) package.

Closing this for now.

PR #3, opened 2016-12-04
Code reorganization

srush:
Made a couple big changes: 

1) No more `args`. Encoder/Decoder now take in already created input_network and output networks and a few explicit named arguments. I like this better because it is self-documenting, and forces a little pain for adding more arguments/

2) Encoder/Decoder no longer know about LSTMs. They know rnn size and num of effective layers. The rest is pushed to the LSTM class. 

3) Encoder/Decoder no longer know anything about words or features. They use the Data class to get inputs, they use a generator to get outputs. Features just become a separate input network and a separate generator. 

4) Small: Encoder backward returns gradients. 

The downside of this PR is that it pushes some logic to `train.lua`. However I would rather try to keep encoder / decoder clean and simple. 

Test plan: Running Integration test.


guillaumekln:
This looks good to me.

I isolated the model creation to a new file, did some minor cleaning and revised the model saving and loading part.

@srush If this also looks good to you, we can merge this. I'm doing more tests but the training logic should be unchanged.

PR #6, opened 2016-12-14
Im2markup

da03:


srush:
Think this merge is a mistake, we're fixing.

da03:
Sorry for that, I mistakenly pushed to the OpenNMT repo instead of my forked one. Already reverted back.

PR #7, opened 2016-12-15
improve adaptivity to im2markup

da03:
Serveral changes to make the onmt code generalizable to the im2markup project (and beyond). All changes shall not affect the existing application. (yet to be tested)
1. Let BiEncoder return gradInput during backward(). (onmt/modules/BiEncoder.lua)
2. Let Batch support double features. Currently the source features are word ids (2D integer tensor of size (sequence_length, batch_size)), in this patch we add interface setSourceInput(sourceInput) to support general double features, e.g. 3D double tensor of size (sequence_length, batch_size, feature_dim). Besides, we add interface setTargetInput, setTargetOutput to directly set targets to provide more flexibility. (onmt/data/batch.lua).
3. Add default zero state gradients for BiEncoder backward(), default zero states for Decoder forward(). This makes it easier for models that do not initialize the decoder states with encoder final states. (onmt/modules/BiEncoder.lua, onmt/modules/Decoder.lua).


onmt/modules/BiEncoder.lua
```diff
@@ -183,5 +186,11 @@ function BiEncoder:backward(batch, gradStatesOutput, gradContextOutput)
     gradContextBwd[{{}, t}]:copy(gradContextOutputBwd[{{}, batch.sourceLength - t + 1}])
   end
 
-  self.bwd:backward(batch, gradStatesOutputBwd, gradContextBwd)
+  local gradInputBwd = self.bwd:backward(batch, gradStatesOutputBwd, gradContextBwd)
+  -- gradInput
+  local gradInput = {}
+  for t = 1, batch.sourceLength do
+      gradInput[t] = torch.add(gradInputFwd[t], gradInputBwd[batch.sourceLength-t+1], 2)
```
guillaumekln: This is quite minor but can we avoid `torch.add` as it allocates new memory? You could add to `gradInputFwd` and return it.

da03: OK will change and update later.

guillaumekln:
Seems like recent changes on `master` introduced a conflict here. As it was just renaming, it should be easy to fix.

da03:
@guillaumekln When you mentioned "when you list arguments on multiple lines, they should ideally be aligned.", where did you refer?

srush:
For future reference, this change should not have been merged. Does not match comment style etc. Let's be strict about this.

da03:
Sure. Should I use another pr to fix the style?

srush:
No it's okay, I just want to be more vigilant about not letting in PRs that
violate the standard.

PR #14, opened 2016-12-20
Update Translator.lua

yhgon:
delete redundant lines in encoder part


guillaumekln:
The lines you deleted are not redundant. They are for loading the decoder.

Were you trying to achieve something else?

srush:
(think this was just an honest mistake)

PR #29, opened 2016-12-23
Support preprocessing larger corpus

guillaumekln:
These changes introduce the use of [`torch.tds`](https://github.com/torch/tds)'s vectors instead of Lua tables during the preprocessing. These data structures do not rely on the Lua memory allocator and thus, prevent the process to reach Lua's memory limit. Additionally, it just impacts the data creation code as `tds.Vec` has the same interface as a standard Lua table.

The small (and acceptable) downside is that it adds a new dependency (torch.tds) which is listed in the README.md.

I validated the change by running the preprocessing on a 16M sentences corpus and initiating a training (past the data loading step) with LuaJIT.


guillaumekln:
It does not break compatibility with already preprocessed data. It is a very internal change.

PR #31, opened 2016-12-23
Split the onmt.utils.Memory.optimize function to generalize to other cases

da03:
The optimize function consists of two parts: one for initializing the output and gradInput tensors, and the other for marking the tensors that can be shared. In this patch, these two parts are separated from each other, while the functionality of onmt.utils.Memory.optimize remains. An outline of the updated optimize function is:
1. model.forward() and model.backward() to initialize the output and gradInput tensors
2. mark shared tensors given an initialized model. This is encapsulated in a new function onmt.utils.Memory.enableMemorySharing(modules), where modules represent a list of modules that we want to optimize. These modules must have been initialized by forward() and backward() beforehand. This function can be used by general cases, e.g., when we have multiple encoders in a new model.


da03:
Still in progress

da03:
Correction: in order to support modules like nn.CMulTable, we need to record additional information like input and gradOutput, which requires us overwriting the original forward and backward function before initializing the tensors.

Usage:
```
local memoryOptimizer = onmt.utils.MemoryOptimizer.new(model)
model:forward(...)
model:backward(...)
memoryOptimizer:optimize()
```

PR #42, opened 2016-12-29
[WIP] Pull out all logic from train.lua

srush:
This code pulls out most of the logic from train.lua. I'm still testing this so don't merge.

1) Primary concern is that a lot of this code is reusable by extensions, but currently it is not in the library part of the onmt, and therefore you end up copying it. 
2) Secondary a lot of the library functions get passed `opt` objects, but they haven't explicitly declared what they are expecting. This gives the classes control of these options. 
3) I still find the train code very hard to read. would like that to be a bit more simplified and documented. 

Future PR will try to do (3). For instance, most of this code should be reusable without it knowing about MT-specific things. 

@da03 Can you confirm that this would be usable in im2text?

### DIFF
--- a/onmt/Models.lua
+++ b/onmt/Models.lua
@@ -1,3 +1,31 @@
+
+local function declareOpts(cmd)
+  cmd:text("")
+  cmd:text("**Model options**")
+  cmd:text("")
+  
+  cmd:option('-layers', 2, [[Number of layers in the LSTM encoder/decoder]])
+  cmd:option('-rnn_size', 500, [[Size of LSTM hidden states]])
+  cmd:option('-word_vec_size', 500, [[Word embedding sizes]])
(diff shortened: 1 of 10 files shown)


da03:
Cool! Hopefully onmt.train.Train can remove all MT specific stuff. For example, I think model shall be a class, and the forward and backward function shall be implemented in model to avoid calling
```local encStates, context = _G.model.encoder:forward(_G.batch)
      local decOutputs = _G.model.decoder:forward(_G.batch, encStates, context)```. This should be something like `model:forward(_G.batch)`, then for other apps, we can use a different model.

da03:
I think so.

guillaumekln:
I'm discarding this PR in favor of #100 that is based on a more recent code base.

PR #43, opened 2016-12-29
Expose the internal translation functions

srush:
Going to expose these functions so that I can make a expert level debug mode for the server.


guillaumekln:
I think a better change would be to make `Translator` a class. It should already be one IMHO. See if #45 works for you.

PR #45, opened 2016-12-30
Define Translator class

guillaumekln:
Make existing Translator component a class.


guillaumekln:
It depends I would say. Some files in `utils` are not classes but collection of functions (`String`, `Table`, etc). Others like `Models` could indeed be a class as @da03 proposed.

PR #57, opened 2017-01-09
WIP: first recipe for EN-FR baseline

vince62s:
@srush Have a look at this.
this is the kind of recipe that makes the system easy to use for newbies (and more).

it's very preliminary but in the same code style as what we do for Kaldi.

Cheers,
Vincent

### DIFF
--- a/egs/baseline-1M-enfr/onmt
+++ b/egs/baseline-1M-enfr/onmt
@@ -1,0 +1,1 @@
+/home/moses/torchwork/OpenNMT/onmt
--- a/egs/baseline-1M-enfr/onmt
+++ b/egs/baseline-1M-enfr/onmt
@@ -1,1 +1,1 @@
-/home/moses/torchwork/OpenNMT/onmt
+../../onmt
--- a/egs/baseline-1M-enfr/newstest/newstest2014-fren-ref.fr.sgm
+++ b/egs/baseline-1M-enfr/newstest/newstest2014-fren-ref.fr.sgm
@@ -1,0 +1,3357 @@
+<refset trglang="fr" setid="newstest2014" srclang="any">
+<doc sysid="ref" docid="00a4704ad22385c64d95064c617f25e4" genre="news" origlang="xx">
+<seg id="1">Les avionneurs se querellent au sujet de la largeur des sièges alors que de grosses commandes sont en jeu</seg>
+<seg id="2">La dispute fait rage entre les grands constructeurs aéronautiques à propos de la largeur des sièges de la classe touriste sur les vols long-courriers, ouvrant la voie à une confrontation amère lors du salon aéronautique de Dubaï qui a lieu de mois-ci.</seg>
+<seg id="3">Le conflit porte sur la largeur des sièges proposés sur les vols long-courriers aux passagers de la classe économique – qui ne sont pas toujours les plus courtisés par les compagnies aériennes, mais auxquels l'espace alloué est essentiel pour augmenter les gains d'efficacité dans les derniers appareils présentés par Airbus SAS et Boeing Co.</seg>
+<seg id="4">Cette semaine, Airbus a appelé l'industrie aéronautique à mettre en place une norme imposant une taille de siège d'au moins 18 pouces (45,72 cm) dans les classes économiques, mais son grand rival américain Boeing déclare que ce devrait être aux compagnies aériennes de décider.</seg>
+<seg id="5">Le différend a éclaté alors que les avionneurs cherchent à vendre des versions encore plus grandes de leurs avions long-courriers bimoteurs, en espérant éventuellement un record de commandes lors de l'évènement qui se déroulera du 17 au 21 novembre.</seg>
+<seg id="6">La façon dont les sièges seront disposés à l'arrière de l'avion – en particulier, y aura-t-il 9 ou 10 sièges de front – est essentielle pour le rendement économique du segment des « mini-jumbo ».</seg>
+<seg id="7">Boeing déclare que son « 777X » réaménagé pourra accueillir en classe économique 406 personnes dans des sièges de plus de 17 pouces de large, et sera configuré avec 10 sièges par rangée.</seg>
+<seg id="8">Airbus indique que la version concurrente de son A350 transportera en classe économique 350 personnes dans des sièges de 18 pouces de large configurés en rangées de 9.</seg>
+<seg id="9">Les géants de l'aéronautique échangent souvent des coups sur des questions techniques en faisant de la publicité dans la presse professionnelle.</seg>
+<seg id="10">Aujourd'hui, Airbus en appelle directement au public avant le salon aéronautique de Dubaï, où le 777X devrait prendre le pas sur ses concurrents avec plus de 100 commandes.</seg>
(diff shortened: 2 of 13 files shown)


egs/baseline-1M-enfr/onmt
```diff
@@ -0,0 +1 @@
+/home/moses/torchwork/OpenNMT/onmt
```
srush: What is this path?

srush:
Thanks for doing this, this will be quite helpful. 

There are couple policy things to decide: 

1) There has been uncertainty in the past about including any Perl code in the repo. I am quite sensitive to this argument. The assertion is that we don't need these scripts of Kaldi/Moses anymore since our interface is significantly more intuitive.
2) Obviously this doesn't apply when the input data comes in XML form.  Therefore, we should probably have an SGM-to-txt converter. This is something we can maybe do in Lua or Python for supportability.
3) There is a stronger argument for including multi-bleu, which is standard and just makes things more convenient.  

Finally I also am not sure these recipes need to be in the main repo. Potentially we could have a OpenNMT/recipes which collects them and includes tutorials? This keeps the surface area of the code base smaller. 

Let me know if you want to talk more in gitter.

vince62s:
oh all these links were supposed to be relative links to OpenNMT/ files or folders, just to be able to execute lua scripts from the recipe-name folders.

Anyway all what you outline above makes sense especially if you want to keep it a pure lua / core code repo.

my view is the following:
For a framework (like tensorflow) it make sense to make these things apart.
For a "toolkit" I think it's good to expose some pre-written scripts like EMS for moses.
it helps users to better adopt the product.

Last, unless I am mistaken, WM-like stuff rely more on sgml format for test sets.

srush:
This is sensible. And I agree short term, having evaluation and WMT SGML
support is crucial. I raised it with the other developers so we can decide
on a policy going forward.

jsenellart-systran:
@vince62s, this is great! fyi, we are preparing a benchmark platform that we will open *very* soon, and the outcome of this recipes will be the baseline system. thanks a lot. I like the idea of recipe and we should commit that in a recipe/ folder (isn't it more explicit than "egs"?). For the evaluation - what I propose (and we are working on that), is that we get all reference testsets we know of, and push them into a reference_testset_xxyy.tgz package on the s3 bucket together with all the tool needed to wrap/unwrap and compute the score. So your recipe will pull the package, and just iterate on all the bash script test_something.sh, to compute all the scores. what do you think?

jsenellart-systran:
new thoughts on that : I do not like starting introducing a lot of new stuff inside the OpenNMT directory - to keep it small/restricted to the minimum. Also being consistent on lua allows us to run complete process wherever we want (like embedded, with minimal dependencies). Why don't we create 2 new repositories: "recipes" (or egs) that  will contain end-to-end scripts, and "3rdParty" that will contain minimal number of tools maintained in other projects like  mteval/multi-bleu? and the tools for sgm conversion could be in OpenNMT/tools - in lua.

vince62s:
No pb to me. There is only one limitation: lua does not accept execution outside the current directory. So I had to create relative symlink to make things "easy for a newcomer" to understand the logic. If we copy this outside the main repo you will have to ask the user to 1) either make the symlink 2) respect the folders name of the 2 repos on his local computer.
Hope this is clear, but I do understand the will to keep the house tight.

guillaumekln:
What about a separate repository but include it as a submodule in OpenNMT?

jsenellart-systran:
after discussion with @guillaumekln - here is what we propose:
- we create a new repository `recipes` that will be a sub-module of OpenNMT. Like Kaldi, I propose that actual recipes are stored in `recipes/name/` directory
- for conveniency, we also create another repo `tools/3rdParty` - where we can commit reference eval scripts and misc conversion scripts useful for the test
- finally, I propose that the actual official test sets license-free that we can collect are stored in s3 bucket (since they won't be recipe-dependent)

@srush, @vince62s - is this ok for you?

vince62s:
Sounds good to me. We'll just have to document the fact that the "recipes" repo shall not be cloned on its own but only used as a submodule.

srush:
Sorry, I'm going to object to the use of submodules. It adds weight to the checkout and complicates things. My proposal is that we should have `tools/thirdparty` checked into the core opennmt repo when the license allows (as it is lightweight and I don't really understand the lua only thing) and `recipes` as a separate repo. We should come up with a technical solution to the lua issue by either making a symlink in the `recipes`, by setting LUAPATH, or by installing opennmt through luarocks.

vince62s:
Guys I am not a difficult person :) - just keep in mind that these "recipes" are supposed to be very light weight. Mainly ONE script, ONE results file and symlinks.
the "data" / "exp" folders are created on the fly on the user local repo and the "newstest" is going away since we will download files from S3 and put them in "data".
for "third party-tools" maybe could stay here as a temporary solution until someone re-writes them in lua.

jsenellart-systran:
I will also object a bit here :) - agree the submodule is not really necessary but starting introducing third party tools inside the code base is dangerous - the two ones we are mentioning are tiny, but there are others that are far bigger around and also multi-bleu and mteval are LGPL not MIT. Finally they are available on a github repository, so the best is that the recipes get them directly from where they are.
So I think - we agree on the recipes repo (I am creating the repo), recipes should get what they need from where they are (opennmt, but also im2text), and for simplifying life of users, we are gathering all known official test sets on s3 bucket (first draft will be done tomorrow). @vince62s, as discussed, we are also pushing 2M baseline corpus and I will come back to you soon on the benchmark platform.

srush:
Okay, so @jsenellart-systran and I are on the same page then. We host `Recipes` as a repo (actually made one alreads), third-part tools come from another repo. We make it very clear how to get recipes and eval on the website, but OpenNMT remains Lua-only and our support tools.

vince62s:
I think I understand everything except for third party tools.
Do we ask the user to install / download them directly ? do we script the downloads for them ?
if the latter, do we put them locally in a specific folder ?

jsenellart-systran:
I propose that it is part of your download at the beginning of the recipe:
```
wget https://raw.githubusercontent.com/moses-smt/mosesdecoder/master/scripts/generic/mteval-v13a.pl
```

as for Kaldi recipe, I propose that we do not plan any optimization by sharing cross-recipe knowledge, so each occurrence of the recipe will download them (we can just check that we don't have them yet in the current directory - which is more critical for the data)

srush:
I agree with one change. Can we make it so the recipes put all their temp
files in a time stamped directory? It should be easy to erase and start
again.

jsenellart-systran:
@vince62s - see https://s3.amazonaws.com/opennmt-tests/onmt_enfr_tests.tgz - we are preparing this type of repository for all languages - this contains all official test sets (older and more recent) from WMT - and we will also include other competitions where testsets are public. this is just a first draft - we do need to add some information about each file in the repo, but the idea will be this one, where we also normalize all the file format for simple processing. it would be great if you can publish the first recipe using these test sets.

vince62s:
will do. I just to make sure I full understand the reported Bleu as per the tutorials.
Usually we do report bleu on detokenized text.

srush:
@vince62s You have write access on `OpenNMT/Recipes`, so feel free to just add these as you go. And we decided to the not import the metrics. Therefore I am going to close this PR as completed.

PR #62, opened 2017-01-10
Use Logger

da03:
Replace print with Logger, which allows printing timestamps and log levels, as well as produces warnings with options if the log file already exists. The two options added are: `-log_file`, which specifies the path to the log file, (if left blank, output to stdout); and `-disable_logs`, which would suppress all outputs.

### DIFF
--- a/onmt/Models.lua
+++ b/onmt/Models.lua
@@ -79,7 +79,13 @@ local function buildDecoder(opt, dicts, verbose)
 
   if opt.input_feed == 1 then
     if verbose then
-      print(" * using input feeding")
+      local log
+      if _G.logger then
+        log = function (...) return _G.logger:info(...) end
+      else
+        log = print
+      end
+      log(" * using input feeding")
     end
     inputSize = inputSize + opt.rnn_size
   end
--- a/onmt/Models.lua
+++ b/onmt/Models.lua
@@ -79,13 +79,7 @@ local function buildDecoder(opt, dicts, verbose)
 
   if opt.input_feed == 1 then
     if verbose then
-      local log
-      if _G.logger then
-        log = function (...) return _G.logger:info(...) end
-      else
-        log = print
-      end
-      log(" * using input feeding")
+      _G.logger:info(" * using input feeding")
     end
     inputSize = inputSize + opt.rnn_size
   end
--- a/onmt/utils/Logger.lua
+++ b/onmt/utils/Logger.lua
@@ -43,6 +43,7 @@ function Logger:__init(logPath, mute)
   else
     self.logFile = nil
   end
+  self.LEVELS = { DEBUG = 0, INFO = 1, WARNING = 2, ERROR = 3 }
 end
 
 --[[ Log a message at a specified level.
@@ -56,10 +57,10 @@ function Logger:log(message, level)
   level = level or 'INFO'
   local timeStamp = os.date('%x %X')
   local msgFormatted = string.format('[%s %s] %s', timeStamp, level, message)
-  if not self.mute then
+  if (not self.mute) and self:_isVisible(level) then
     print (msgFormatted)
   end
-  if self.logFile then
+  if self.logFile and self:_isVisible(level) then
     self.logFile:write(msgFormatted .. '\n')
     self.logFile:flush()
   end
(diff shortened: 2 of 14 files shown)


onmt/Models.lua
```diff
@@ -79,7 +79,13 @@ local function buildDecoder(opt, dicts, verbose)
 
   if opt.input_feed == 1 then
     if verbose then
-      print(" * using input feeding")
+      local log
+      if _G.logger then
```
guillaumekln: Why do we need to make this conditional branching?

preprocess.lua
```diff
@@ -32,6 +32,8 @@ cmd:option('-shuffle', 1, [[Shuffle data]])
 cmd:option('-seed', 3435, [[Random seed]])
 
 cmd:option('-report_every', 100000, [[Report status every this many sentences]])
+cmd:option('-log_file', '', [[Outputs logs to a file under this path instead of stdout]])
```
guillaumekln: I think we should apply the pattern introduced by @srush: defining a static function `declareOpts` in the Logger that declares its options.

See for example with the Translator:

* https://github.com/OpenNMT/OpenNMT/blob/master/onmt/translate/Translator.lua#L3
* https://github.com/OpenNMT/OpenNMT/blob/master/translate.lua#L20

preprocess.lua
```diff
@@ -275,6 +277,12 @@ local function main()
 
   onmt.utils.Opt.init(opt, requiredOptions)
 
+  local mute = (opt.log_file:len() > 0)
```
guillaumekln: This should ideally be factorized in a static `init` function inside the Logger which returns a new Logger object depending on the options.

da03:
Modifying, do not merge

da03:
Okay, ready to be reviewed.

da03:
Okay ready to be reviewed.

onmt/utils/Logger.lua
```diff
@@ -93,7 +124,61 @@ Parameters:
...
+  end
+end
+
+--[[ Set the visible message level. Lower level messages will be muted.
+
+Parameters:
+  * `level` - 'DEBUG', 'INFO', 'WARNING' or 'ERROR'.
+
+]]
+function Logger:setVisibleLevel(level)
```
guillaumekln: While you are at it, can you add a `-log_level` option so that the user can control this?

train.lua
```diff
@@ -486,14 +487,14 @@ local function main()
...
-    print(string.format(' * vocabulary size: source = %d; target = %d',
-                        dataset.dicts.src.words:size(), dataset.dicts.tgt.words:size()))
-    print(string.format(' * additional features: source = %d; target = %d',
-                        #dataset.dicts.src.features, #dataset.dicts.tgt.features))
-    print(string.format(' * maximum sequence length: source = %d; target = %d',
-                        trainData.maxSourceLength, trainData.maxTargetLength))
-    print(string.format(' * number of training sentences: %d', #trainData.src))
-    print(string.format(' * maximum batch size: %d', opt.max_batch_size))
+    _G.logger:info(' * vocabulary size: source = %d; target = %d',
+                        dataset.dicts.src.words:size(), dataset.dicts.tgt.words:size())
```
guillaumekln: Can you align the second lines with the opening parenthesis? Like before:

```lua
print(string.format(' * vocabulary size: source = %d; target = %d',
                    dataset.dicts.src.words:size(), dataset.dicts.tgt.words:size()))
```

translate.lua
```diff
@@ -26,11 +26,12 @@ cmd:option('-gpuid', 0, [[1-based identifier of the GPU to use. CPU is used when
...
 cmd:option('-time', false, [[Measure batch translation time]])
 
+onmt.utils.Logger.declareOpts(cmd)
 
 local function reportScore(name, scoreTotal, wordsTotal)
-  print(string.format(name .. " AVG SCORE: %.4f, " .. name .. " PPL: %.4f",
+  _G.logger:info(name .. " AVG SCORE: %.4f, " .. name .. " PPL: %.4f",
                       scoreTotal / wordsTotal,
-                      math.exp(-scoreTotal/wordsTotal)))
+                      math.exp(-scoreTotal/wordsTotal))
```
guillaumekln: Align with the parenthesis these two lines also.

onmt/utils/Logger.lua
```diff
@@ -2,20 +2,50 @@
...
+
+    local cmd = torch.CmdLine()
+    onmt.utils.Logger.declareOpts(cmd)
+    opt = cmd:parse(arg)
+    logger = onmt.utils.Logger.new(opt)
+    logger:info('%s is an extension of OpenNMT.', 'Im2Text')
+    logger:shutDown()
+
+]]
+function Logger:__init(args)
```
guillaumekln: As there are few options, maybe you can pass them explicitly.

```lua
function Logger:__init(logFile, disableLogs, logLevel)
```

da03:
okay ready to go

PR #66, opened 2017-01-11
Catching last batch when creating minibatch offsets

swiseman:
Catches last batch when making mini-batch offsets in onmt/data/Dataset.lua


guillaumekln:
Good catch! Literally.

srush:
Accepting. Please use full sentences (with periods) for comments.

swiseman:
Got it, thanks.

PR #67, opened 2017-01-11
Clone timesteps with clone()

swiseman:
Creates per-timestep clones in onmt/modules/Sequencer.lua with network:clone(), rather than by writing to disk.

### DIFF
--- a/onmt/modules/Sequencer.lua
+++ b/onmt/modules/Sequencer.lua
@@ -24,21 +24,7 @@ function Sequencer:__init(network)
 end
 
 function Sequencer:_sharedClone()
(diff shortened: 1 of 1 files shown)


guillaumekln:
Thanks! I agree it is simpler and it will be merged.

Is it really faster though? The [`clone` implementation](https://github.com/torch/nn/blob/master/Module.lua#L133) also serializes modules into a `MemoryFile` (which is not on disk by the way) then shares the parameters.

swiseman:
Cool! Interesting, yeah, in my experience it does lead to a noticeable speedup (esp. if there are very long sequences in the dataset), but I haven't extensively tested it in all situations.

PR #82, opened 2017-01-19
Fix #81

### LINKED ISSUE #81: Log error occured in training start
/root/torch/install/bin/luajit: ./onmt/utils/Log.lua:7: attempt to call method 'write' (a nil value)
stack traceback:
        ./onmt/utils/Log.lua:7: in function 'logJsonRecursive'
        ./onmt/utils/Log.lua:27: in function 'logJson'
        train.lua:516: in function 'main'
        train.lua:554: in main chunk
        [C]: in function 'dofile'
        /root/torch/install/lib/luarocks/rocks/trepl/scm-1/bin/th:145: in main chunk
        [C]: at 0x00406670

da03:
typo in Logger.lua

### DIFF
--- a/onmt/utils/Log.lua
+++ b/onmt/utils/Log.lua
@@ -1,31 +1,31 @@
 local function logJsonRecursive(obj)
   if type(obj) == 'string' then
-    _G.logger:write('"' .. obj .. '"')
+    _G.logger:writeMsg('"' .. obj .. '"')
   elseif type(obj) == 'table' then
     local first = true
 
-    _G.logger:write('{')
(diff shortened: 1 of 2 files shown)


guillaumekln:
Sorry, I changed `write` to `writeMsg` without updating the calling codes.

However, we have to avoid `write` for a good (and interesting) reason. In parallel mode, each thread has its own context so we must share the global `Logger` instance. To move this instance inside each context, Torch uses serialization and attempts to call `write` on the object. But here `write` does not serialize the object and it fails.

I updated the calling codes in your branch.

PR #91, opened 2017-01-23
fix allow memory access between devices

voidmagic:
Allow memory access between devices should be cutorch.setKernelPeerToPeerAccess(true), not cutorch.getKernelPeerToPeerAccess(true), according to the cutorch document.

### DIFF
--- a/onmt/utils/Cuda.lua
+++ b/onmt/utils/Cuda.lua
@@ -35,7 +35,7 @@ function Cuda.init(opt, masterGPU)
       _G.logger:info('Using GPU(s): ' .. table.concat(Cuda.gpuIds, ', '))
 
       -- Allow memory access between devices.
-      cutorch.getKernelPeerToPeerAccess(true)
+      cutorch.setKernelPeerToPeerAccess(true)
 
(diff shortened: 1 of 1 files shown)


guillaumekln:
Thank you, the typo was always there it seems.

@jsenellart-systran Do we actually want to enable this flag?

From `cutorch` documentation:

> `cutorch.setKernelPeerToPeerAccess(f)`: by default, kernels running on one device cannot directly access memory on another device. This is a check imposed by cutorch, to prevent synchronization and performance issues. To disable the check, call this with `f` true. Kernel p2p access is actually only allowed for a pair of devices if both this is true and the underlying `getPeerToPeerAccess` for the pair involved is true.

jsenellart-systran:
actually we don't want it. in multi-gpu mode, the only operation authorized without the flag set and used by the code is the copy operation. Any other operation has a huge performance cost.

guillaumekln:
@voidmagic Could you then simply delete the line so that we can merge your patch?

voidmagic:
OK, I pulled a new requests.

PR #93, opened 2017-01-23
[WIP] Introduce CUDNN support

jsenellart-systran:
enabled with: 
`-cudnn m1,m2` option where `mi` in Activation, SoftMax, RNN enabling respectively cudnn for activation functions, for SoftMax layers (should be refined cause there is a "small" softmax layer in attention, and a very large in generation), and RNN.

*won't work* with residual layers and brnn merge

WIP:
- missing the conversion from cudnn to non cudnn parameters.
- some work to be done on bidirectional CUDNN
- more benchmark especially for LogSoftMax


jsenellart-systran:
closing this PR which now replaced by #118.

PR #95, opened 2017-01-25
Support individually setting features vocabulary and embedding size

guillaumekln:
The existing word features framework is mostly targeted to features with few values (case, POS, etc.) with no way to limit the vocabulary size of the feature or set an appropriate embedding size. 

These *non-breaking* changes remove these limitations and also bring:

* a more generic `FeaturesEmbeddings` module that no more relies on the `Dict` class.
* the ability to use a different word embedding size for the encoder and decoder.

See http://forum.opennmt.net/t/solved-multi-way-embeddings/138.

---

@srush Could you take a quick look at this PR and tell if the changes look reasonable to you?


jsenellart-systran:
for me it is ok - we will need to very soon introduce more checks and consistencies on the values of these parameters that have more structure (and probably also for all of them). in `-adaptive_softmax` - I used a notation like `{id1,id2}`

guillaumekln:
Why did you use brackets? It seems a bit risky as it has a meaning in `bash`:

```
$ echo {ye,ee}
ye ee
$ echo b{ye,ee}
bye bee
```

while comma-separated strings have not:

```
$ echo aa,ee
aa,ee
```

jsenellart-systran:
I agree! my point is that we do need to start normalizing :) and adding some validation mechanism

PR #97, opened 2017-01-27
Add FP16 support

jsenellart-systran:
FP16 support is complete and tested like that but it does need fix for https://github.com/torch/cunn/issues/417 and https://github.com/torch/cunn/issues/380.


guillaumekln:
The option will work when https://github.com/torch/cutorch/pull/720 will be merged in cutorch.

PR #101, opened 2017-01-30
Added RecurrentBatchNorm module

rachtsingh:
This PR adds support for recurrent batch normalization to the LSTM and GRU implementations, via the flag `-batch_norm`.

The most important structural change is that the new RecurrentBatchNorm module contains inside it `nn.BatchNormalization` modules, one for each time step. We select the correct one to use in forward/backward passes inside the `:net` function inside `onmt.Sequencer`. This makes the function have side effects, though it should be transparent from the outside.

I'm not 100% sure the implementation for GRU is correct, but the LSTM one does perform better with batch norm on.


onmt/modules/RecurrentBatchNorm.lua
```diff
@@ -0,0 +1,91 @@
...
+  self.momentum = momentum
+  self.affine = affine
+  self.noBias = noBias
+
+  self:setTimeStep(1)
+end
+
+function RecurrentBatchNorm:__timeStepClone()
+  local bn = nn.BatchNormalization(self.dim, self.eps, self.momentum, self.affine)
+  if self._type == 'torch.CudaTensor' then
```
guillaumekln: It is more generic to do:

```lua
bn:type(self._type)
```

And it will also fix CPU execution as we are working with `float`.

onmt/modules/Sequencer.lua
```diff
@@ -26,6 +28,16 @@ end
 function Sequencer:_sharedClone()
   local clone = self.network:clone('weight', 'gradWeight', 'bias', 'gradBias')
 
+  local cloneRNN = onmt.utils.Tensor.findModule(clone, self.rnnClass)
+  if cloneRNN ~= nil then
+    cloneRNN:shareRecurrents(onmt.utils.Tensor.findModule(self.network, self.rnnClass))
+  end
+  function clone:setTimeStep(t)
```
guillaumekln: I think it would be better to just build an array of all RNNs (in `self.rnns`) and define this function statically in the `Sequencer` class.

Also `_sharedClone` is not called at test time so `Sequencer.setTimeStep` will be undefined.

onmt/modules/Sequencer.lua
```diff
@@ -26,6 +28,16 @@ end
 function Sequencer:_sharedClone()
   local clone = self.network:clone('weight', 'gradWeight', 'bias', 'gradBias')
 
+  local cloneRNN = onmt.utils.Tensor.findModule(clone, self.rnnClass)
```
rachtsingh: This is a much better solution! I didn't read the code carefully enough to see this is what was happening. Fixed.

onmt/modules/LSTM.lua
```diff
@@ -104,8 +116,22 @@ function LSTM:_buildLayer(inputSize, hiddenSize)
   local x = inputs[3]
 
   -- Evaluate the input sums at once for efficiency.
-  local i2h = nn.Linear(inputSize, 4 * hiddenSize)(x)
-  local h2h = nn.Linear(hiddenSize, 4 * hiddenSize)(prevH)
+  local i2h = nn.Linear(inputSize, 4 * hiddenSize, false)(x)
```
rachtsingh: So first, in the non-batch-normalized case, by having both modules have biases, there's an unnecessary redundancy since they will just be added anyway.

In the batch normalized case, the authors of this [paper](https://arxiv.org/pdf/1603.09025v4.pdf) recommended batch normalizing the output before adding the bias, and having that be separate (see equation 6).

That all being said, I'll measure the differences and report back.

onmt/modules/RecurrentBatchNorm.lua
rachtsingh: Sounds great! fixed.

onmt/modules/Sequencer.lua
rachtsingh: Ok, I thought that was a little hacky but if its acceptable it's definitely a better solution. 

I overlooked the second fact - that might be what's leading to your NaN perplexity on validation. I will fix this in a second and then push to this PR.

rachtsingh:
Here's a log from running the baseline: https://gist.github.com/rachtsingh/ff6c22a6dfadb41808e9586f8d905163

And the equivalent without batch normalization:
https://gist.github.com/rachtsingh/bdd5712047171f4944c75af01a80af2c

So the validation perplexity doesn't fall faster, but training does. In an earlier experiment it did lead to a 2x smaller validation perplexity after the first epoch, so it might be situation dependent. 

In any case the wall clock speed is much worse, and I'm not entirely sure why - there's a 5% increase in model size, but a 2x decrease in speed?

onmt/modules/LSTM.lua
guillaumekln: cc @jsenellart-systran

jsenellart-systran: for reference, we know that there is a redundancy here - but we added it for being equivalent with cudnn.LSTM. When batch-normalization will be used - the module will not be anymore compatible with cudnn.LSTM anyway, so in that case you can make the change, but let us manage to keep the non-batch-normalization mode as is.

onmt/modules/RecurrentBatchNorm.lua
```diff
@@ -0,0 +1,89 @@
...
+
+  self:setTimeStep(1)
+end
+
+function RecurrentBatchNorm:__timeStepClone()
+  local bn = nn.BatchNormalization(self.dim, self.eps, self.momentum, self.affine)
+  bn:type(self._type)
+  if self.affine then
+    -- initialize weights using the recommended value
+    bn.weight:fill(0.1)
```
guillaumekln: From the paper:

> Note that we separate only the statistics over time and **not the γ and β parameters**.

You may need to call `bn:share(self.modules[1], 'weight', 'gradWeight', 'bias', 'gradBias')` for `t > 1`.

onmt/modules/LSTM.lua
guillaumekln: So we prefer `not self.batchNorm` instead of `false`.

onmt/utils/MemoryOptimizer.lua
```diff
@@ -14,7 +14,9 @@ local protectInput = {
   'nn.Linear',
   'nn.JoinTable',
   'nn.CMulTable',
-  'nn.MM'
+  'nn.MM',
+  'nn.BatchNormalization',
+  'onmt.RecurrentBatchNorm',
```
guillaumekln: Just adding `nn.BatchNormalization` is enough as they have both the same input.

guillaumekln:
Did you make more experiments with it? While the implementation seems correct (with parameters sharing), I can't get it to work on a baseline corpus.

Validation perplexity is high, going up and down and predictions are random.

PR #111, opened 2017-02-07
Use LongTensor to fit nn.LookupTable

anoidgit:
If you use the other types except LongTensor, the nn.LookupTable will do a convert job every time before it forward.


anoidgit:
https://github.com/torch/nn/blob/master/LookupTable.lua#L58
This may need much more disk space, but I do not think It will be a problem.

guillaumekln:
It is right that the input is copied before looking up the embedding due to type mismatch (assuming the input is always contiguous otherwise it will be copied anyway).

However, the change is made at the wrong place. You want to change tensor types when batches are created and not when data are prepared (where we absolutely don't want to double the memory requirements!).

See:

* https://github.com/OpenNMT/OpenNMT/blob/master/onmt/data/Batch.lua#L73 for the source sequence
* https://github.com/OpenNMT/OpenNMT/blob/master/onmt/data/Batch.lua#L90 for the target sequence

Also, additional work is needed for the GPU because only calling `cuda()` is no more sufficient:
```
th> torch.LongTensor():cuda()
[torch.CudaTensor with no dimension]
```

We need to use `cudaLong()` instead:
```
th> torch.LongTensor():cudaLong()
[torch.CudaLongTensor with no dimension]
```

See the function [`Cuda.convert`](https://github.com/OpenNMT/OpenNMT/blob/master/onmt/utils/Cuda.lua#L54). We should then replace `return obj:cuda()` by something like:

```lua
if torch.typename(obj):find('torch%..*LongTensor') then
  return obj:cudaLong()
else
  return obj:cuda()
end
```

anoidgit:
You are right, and I'M not familiar with the project structure, I'm trying to change my code with your advice, but I think there are some reason that nn.LookupTable use LongTensor as its input.

onmt/utils/Cuda.lua
```diff
@@ -52,18 +52,23 @@ end
...
     elseif not Cuda.activated and obj.float ~= nil then
       -- Defaults to float instead of double.
       return obj:float()
     end
-  end
-
-  if torch.typename(obj) or type(obj) == 'table' then
-    for k, v in pairs(obj) do
-      obj[k] = Cuda.convert(v)
+  else
```
guillaumekln: This part of the function does not require any modification. It was correct as is.

anoidgit: I have already make the changes you asked to, but I still think this part of code could be merged to the first part, at least the repeated call to typename is unnecessary.

anoidgit:
I think that https://github.com/OpenNMT/OpenNMT/blob/master/onmt/utils/Cuda.lua#L55 have checked the typename, so a 2nd time check is unnecessary at https://github.com/OpenNMT/OpenNMT/blob/master/onmt/utils/Cuda.lua#L64, so I merged the second part of the function to the first part.

guillaumekln:
It is needed as some objects have a `typename` but do not implement the `cuda()` or `float()` functions. In this case, we still want to convert each of their fields.

onmt/utils/Cuda.lua
```diff
@@ -52,16 +52,21 @@ end
...
     if Cuda.activated and obj.cuda ~= nil then
-      return obj:cuda()
+      if objtype:find('torch%..*LongTensor') then
+        return obj:cudaLong()
+      else
+        return obj:cuda()
+      end
     elseif not Cuda.activated and obj.float ~= nil then
       -- Defaults to float instead of double.
       return obj:float()
```
guillaumekln: Oh, we need to do the same way on the back to CPU:

```lua
if objtype:find('torch%..*LongTensor') then
  return obj:long()
else
  return obj:float()
end
```

With this final change we are good to go!

anoidgit: You are so careful, I think I have done the job by changing the condition of elseif. @guillaumekln

onmt/utils/Cuda.lua
```diff
@@ -52,16 +52,21 @@ end
...
+  if objtype then
     if Cuda.activated and obj.cuda ~= nil then
-      return obj:cuda()
-    elseif not Cuda.activated and obj.float ~= nil then
+      if objtype:find('torch%..*LongTensor') then
+        return obj:cudaLong()
+      else
+        return obj:cuda()
+      end
+    elseif not Cuda.activated and obj.float ~= nil and not objtype:find('torch%..*LongTensor') then
```
guillaumekln: Close. :)

`CudaLongTensor` are not moved back to CPU with this test. The piece of code I shared handled this case.

anoidgit: Do you think It  is ok if I just compare the strings? @guillaumekln

guillaumekln: I guess you can, it's just that your code now converts a `CudaLongTensor` back to a `FloatTensor`. With this piece:

```lua
if objtype:find('torch%..*LongTensor') then
  return obj:long()
else
  return obj:float()
end
```

there is a nice symmetry with the GPU part and works as expected. By the way, calling `long()` on a `LongTensor` does nothing.

anoidgit: Thank you for your warning, seems I have some misunderstand with `Cuda.activated` before. @guillaumekln

guillaumekln:
I just fixed a very minor typo and now it looks good. Thank you for your contribution!

anoidgit:
I have learnt many things from you. You guide me for these changes and you are welcome. @guillaumekln

PR #115, opened 2017-02-09
 Scoring translations with BLEU

netxiao:
A simple BLEU scoring tool is the script multi-bleu.perl:
` multi-bleu.perl reference < mt-output`

> Reference file and system output have to be sentence-aligned (line X in the reference file corresponds to line X in the system output). If multiple reference translation exist, these have to be stored in seperate files and named **reference0**, **reference1**, **reference2**, etc. All the texts need to be tokenized.
> 
> A popular script to score translations with BLEU is the [NIST mteval script](http://www.nist.gov/speech/tests/mt/2008/scoring.html). It requires that text is wrapped into a SGML format. This format is used for instance by the [NIST evaluation](http://www.nist.gov/speech/tests/mt/2009/) and the [WMT Shared Task evaluations](http://www.statmt.org/wmt09/translation-task.html). See the latter for more details on using this script.

### DIFF
--- a/benchmark/3rdParty/multi-bleu.perl
+++ b/benchmark/3rdParty/multi-bleu.perl
@@ -1,0 +1,118 @@
+#!/usr/bin/perl -w
+
+use strict;
+
+if (!scalar(@ARGV)) {
+  print STDERR "Syntax: multi-bleu.perl [ref-stem] < [system-output]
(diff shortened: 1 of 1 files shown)


jsenellart-systran:
Thanks!

PR #119, opened 2017-02-09
Caseless BPE

DYCSystran:
- Add support for case_insensitive BPE (-lc for learn_bpe.lua during training and -bpe_case_insensitive for BPE.lua during apply)
- Add learn_bpe.lua with different modes of learning (prefix, suffix, both or none)
- Use presentation form characters for BOT (Begin_Of_Token, ﹤ ) and EOT (﹥), but the former EOT can still be used by changing the option EOT_marker
- Add BPE "record" mechanism by adding the possibility of tokenizing directly the raw input in learn_bpe.lua, with the same tokenizer as in tokenize.lua and saving the BPE/tokenization options into the BPE model, which will be automatically read and used in apply mode


tools/utils/learn_bpe.lua
```diff
@@ -0,0 +1,247 @@
+require('torch')
+
+local unicode = require 'tools.utils.unicode'
+local tokenizer = require('tools.utils.tokenizer')
```
jsenellart-systran: this file should be directly in tools/

tools/utils/learn_bpe.lua
```diff
@@ -0,0 +1,247 @@
...
+cmd:text("**learn_bpe.lua**")
+cmd:text("")
+
+cmd:option('-input', '', [[Input file for bpe learning]])
+cmd:option('-size', '30000', [[The number of merge operations to learn]])
+cmd:option('-t', false, [[tokenize the input with tokenizer, the same options as tokenize.lua, but only '-mode' is taken into account for BPE training]])
+cmd:option('-mode', 'conservative', [[Define how aggressive should the tokenization be - 'aggressive' only keeps sequences of letters/numbers, 'conservative' allows mix of alphanumeric as in: '2,000', 'E65', 'soft-landing']])
+cmd:option('-lc', false, [[lowercase the output from the tokenizer before BPE learning]])
+cmd:option('-prefix', false, [[Append '﹤' to the begining of each word to learn prefix-orientated pair statistics]])
+cmd:option('-suffix', false, [[Append '﹥' to the end of each word to learn suffix-orientated pair statistics]])
```
jsenellart-systran: orien*ted*

tools/utils/learn_bpe.lua
```diff
@@ -0,0 +1,247 @@
...
+local case = require ('tools.utils.case')
+local separators = require('tools.utils.separators')
+
+local cmd = torch.CmdLine()
+
+cmd:text("")
+cmd:text("**learn_bpe.lua**")
+cmd:text("")
+
+cmd:option('-input', '', [[Input file for bpe learning]])
```
jsenellart-systran: `-input` is not consistent with other module - I would use STDIN for input but on the other hand add a `-save_bpe` for saving bpe model

tools/tokenize.lua
```diff
@@ -15,7 +15,12 @@ cmd:option('-joiner_annotate', false, [[Include joiner annotation using 'joiner'
 cmd:option('-joiner', separators.joiner_marker, [[Character used to annotate joiners]])
 cmd:option('-joiner_new', false, [[in joiner_annotate mode, 'joiner' is an independent token]])
 cmd:option('-case_feature', false, [[Generate case feature]])
-cmd:option('-bpe_model', '', [[Apply Byte Pair Encoding if the BPE model path is given]])
+cmd:option('-bpe_model', '', [[Apply Byte Pair Encoding if the BPE model path is given. If the option is used, 'mode' will be overridden/set automatically if the BPE model specified by bpe_model is learnt using learn_bpe.lua]])
+cmd:option('-EOT_marker', separators.EOT, [[Marker used to mark the end of token, use '</w>' for python models, otherwise default value ]])
+cmd:option('-BOT_marker', separators.BOT, [[Marker used to mark the begining of token]])
+cmd:option('-bpe_case_insensitive', false, [[Apply BPE internally in lowercase, but still output the truecase units. This option will be overridden/set automatically if the BPE model specified by bpe_model is learnt using learn_bpe.lua]])
```
DYCSystran: For those who use models learnt by learn_bpe.py on a lowercased training set with their own tokenization

tools/utils/BPE.lua
```diff
@@ -26,8 +42,28 @@ function BPE:__init(codesfile_path, joiner_annotate, joiner_new)
     end
     t=f:read("*line")
   end
-  self.joiner_new = joiner_new
-  self.joiner_annotate = joiner_annotate
+end
+
+local function utf8len (s)
```
DYCSystran: OK

tools/utils/BPE.lua
```diff
@@ -14,8 +13,25 @@ function BPE:__init(codesfile_path, joiner_annotate, joiner_new)
...
+
+  self.EOT_marker = opt.EOT_marker
+  self.BOT_marker = opt.BOT_marker
+  self.joiner_new = opt.joiner_new
+  self.joiner_annotate = opt.joiner_annotate
+
   local t = f:read("*line")
+  local options = self.split(t, ";")
+  if (#options == 4) then
+    self.prefix = options[1] == "true" and true or false
```
jsenellart-systran: thanks Yongchao - I did not ask where it was coming from, but why you were using this tautological expression...

tools/tokenize.lua
DYCSystran: With some side effects by using tow different lowercasers for training time and apply time

tools/utils/BPE.lua
```diff
@@ -38,9 +52,15 @@ local function getPairs(word)
   return pairs
 end
 
-local function str2word(l)
+local function str2word(l, case_insensitive)
```
jsenellart-systran: this function should be in unicode.lua

tools/utils/learn_bpe.lua
```diff
@@ -0,0 +1,272 @@
+require('torch')
```
jsenellart-systran: as discussed - please move to tools/

tools/utils/BPE.lua
DYCSystran: It's a function only used for BPE and does not have any unicode functionality. Is it really necessary to move it into unicode.lua?

tools/utils/learn_bpe.lua
DYCSystran: OK

jsenellart-systran:
Yongchao - do you have side-documentation? are the default training options corresponding to standard BPE?

DYCSystran:
Jean, I only added option descriptions for the moment, I could add some documentation or special use cases on the "advanced guide" page of OpenNMT.

The default training options correpond to the standard BPE, except the fact that the EOT symbol is different (﹥ instead of </w> in the original python script, and I added ﹤ as BOT), so one should use the OpenNMT's tokenize.lua to apply BPE but not the original python apply.

However, it's possible to apply models trained by the original python scripts by specifying -EOT_marker </w> for tokenize.lua

jsenellart-systran:
Yongchao - please add a section in the documentation and ideally a page in the forum to describe in detail what you did

PR #136, opened 2017-02-19
[WIP] More complex encoders

jsenellart-systran:
make encoder structure more generic and introduce:
- dbrnn (deep bidirectional rnn) - which is a stack of bidirectional RNN
- pyramidal dbrnn for https://arxiv.org/pdf/1508.01211.pdf


anoidgit:
Maybe only make the first layer of encoder bidirectional is enough according to GNMT?

jsenellart-systran:
yes - I will add that as an option to the dbrnn.

anoidgit:
That will be awsome. @jsenellart-systran

guillaumekln:
A quick feedback if we plan to merge this in the future. Renaming these fields are breaking existing models:

`SimpleEncoder.lua`:
* `Encoder` → `SimpleEncoder`

`BiEncoder.lua`:
* `self.args.rnnSize` → `self.args.rnn_size`
* `self.args.merge` → `self.args.brnn_merge`

jsenellart-systran:
I merged this PR into #142 and will follow-up from there.
@guillaumekln - for Encoder->SimpleEncoder, I will rename back SimpleEncoder into Encoder for compatibility - for variable change in BiEncoder - I don't think we have a problem.

PR #138, opened 2017-02-20
Add standalone script to generate dictionaries

guillaumekln:


guillaumekln:
@vince62s Let me know if that covers your need. It's a simple script that take a data file and saves dictionaries that can be reused during data preprocessing.

vince62s:
so fast, many thanks.
Ideally, it would be great if we could have several data files (comma separated ?).
Use case would be as follow: we concat a first set of "out of domain" files => tokenize
we concat a second set of "in domain" files => tokenize
we build the vocab based on the 2 tokenized files (avoiding to merge them)
then we preprocess the 2 tokenized files individually
it is not big deal, but would avoid to merge some something we don't need to.
Cheers.

guillaumekln:
I'm not a fan of comma-separated paths on the command line as the shell can't auto-complete past the second one. So I think the user would loose more time setting the files on the command line than doing:

```
cat file1 file2 > files
```

vince62s:
ok fine, he will need to do "rm files" too :)
no big deal. I'll test it so that you can check it in and close.

vince62s:
I get this:
torch/install/bin/luajit: tools/build_dict.lua:3: attempt to index field 'ExtendedCmdLine' (a nil value)
stack traceback:
	tools/build_dict.lua:3: in main chunk
	[C]: in function 'dofile'
	...oses/torch/install/lib/luarocks/rocks/trepl/scm-1/bin/th:150: in main chunk
	[C]: at 0x00406670
/home/moses/torch/install/bin/luajit: tools/build_dict.lua:3: attempt to index field 'ExtendedCmdLine' (a nil value)
stack traceback:
	tools/build_dict.lua:3: in main chunk
	[C]: in function 'dofile'
	...oses/torch/install/lib/luarocks/rocks/trepl/scm-1/bin/th:150: in main chunk
	[C]: at 0x00406670

guillaumekln:
You need the latest code revision to make it work (`ExtendedCmdLine` was recently introduced).

vince62s:
works fine with some corpora, but I am having issues with some others (Globalvoices for instance)

I need a more verbose message on the line I have the error otherwise it's not easy....
Might be similar in preprocess.lua

torch/install/share/lua/5.1/onmt/data/Vocabulary.lua:35: all sentences must have the same numbers of additional features
stack traceback:

guillaumekln:
This means your text contains the word features separator `￨`. Running the tokenization before should fix it, did you run it before the preprocessing to make it work?

vince62s:
it is tokenized. once tokenized, through "preprocess" works fine, through "build_dict" I get the error ....

PR #142, opened 2017-02-23
[WIP] Complex Encoders and vector inputs (used for ASR)

jsenellart-systran:
Changed implementation to separate preprocess logic (feature extraction from training logic).

Added 2 new features for the preprocess:
- option `-idx_files` indicate that source/target files are indexed meaning that their format is `key value` not necessarily in the same order
- new data type `feattext` which is using Kaldi text ark dump format 

```
KEY [
VAL1 VAL2 VAL3
VAL4 VAL5 VAL6 ]
KEY2 [
...
]
```

Typical preprocessing command:
```
th preprocess.lua -data_type 'feattext' -train_src TIMIT/raw_mfcc_train.ark.txt \
                -train_tgt TIMIT/data/train/text -valid_src TIMIT/raw_mfcc_dev.ark.txt \
                -valid_tgt TIMIT/data/dev/text -save_data TIMIT/baseline -idx_files \
                -src_seq_length 1000 -tgt_seq_length 100
```

For training:
```
th train.lua -data ~/Downloads/TIMIT/baseline-train.t7 -save_model ~/Downloads/TIMIT -pdbrnn -report_every 1 -rnn_size 800 -word_vec_size 20  -layers 4 -max_batch_size 16 -learning_rate 0.7 -learning_rate_decay 0.8 -end_epoch 20
```

For decoding:
```
th translate.lua -model data/asr1_epoch15_8.17.t7 -src TIMIT/raw_mfcc_test.ark.txt  -batch_size 1
```

(batch_size 1 is for the moment necessary)

PR #152, opened 2017-03-01
Sampling

jungikim:
Implements sampled dataset.
Sampling can be done via uniform or perplexity-based multinomial probability distribution.

perplexity-based sampling is experimental (effects to be experimented).

When perplexity-based sampling is required, an optional return value indvLoss is passed up from  decoder -> Seq2Seq -> trainer.


jungikim:
As-is, this PR introduces an additional dependency on torchx.
We can avoid it if necessary by only including the necessary module from https://github.com/nicholas-leonard/torchx/blob/master/AliasMultinomial.lua (BSD-3-Clause)

onmt/modules/Decoder.lua
```diff
@@ -329,14 +330,25 @@ function Decoder:backward(batch, outputs, criterion)
...
     -- Compute decoder output gradients.
     -- Note: This would typically be in the forward pass.
     local pred = self.generator:forward(outputs[t])
     local output = batch:getTargetOutput(t)
 
-    loss = loss + criterion:forward(pred, output)
+    if self.args.indvLoss then
+      local curLossSum = 0
+      for i=1, pred[1]:size(1) do
+        local tmpLoss = criterion:forward({pred[1][i]},{output[1][i]})
```
guillaumekln: This does not work with target features. For example with one target feature:

`pred` is in the form:

```
{
  1 : CudaTensor - size: 64x34058
  2 : CudaTensor - size: 64x9
}
```

and `output`:

```
{
  1 : CudaLongTensor - size: 64
  2 : CudaLongTensor - size: 64
}
```

onmt/modules/Decoder.lua
```diff
@@ -365,8 +377,12 @@ function Decoder:backward(batch, outputs, criterion)
       gradStatesInput[i]:copy(gradInput[i])
     end
   end
+  
+  if self.args.indvLoss then
+    indvAvgLoss:div(batch.targetLength)
```
guillaumekln: Shouldn't each loss be normalized by each target true length?

onmt/data/SampledDataset.lua
```diff
@@ -0,0 +1,266 @@
...
+      if self.ppl[i] ~= self.sample_w_ppl_init and self.ppl[i] > threshold then
+        -- asign low value to instances with ppl above threshold (outliers)
+        self.samplingProb[i] = 1
+      else
+        self.samplingProb[i] = self.ppl[i]
+      end
+    end
+  end
+
+--  local sampled = torch.multinomial(self.samplingProb:view(1,#self.src), self.samplingSize, --[[replacement]] false)
```
jungikim: I'd like to keep this comment even if it is dead code just as a reminder of how it is done by generic torch library. Maybe the torch function will be updated for speed improvement in the future.

onmt/data/SampledDataset.lua
```diff
@@ -0,0 +1,266 @@
+--[[ Data management and batch creation. Handles data created by `preprocess.lua`. ]]
+require 'torchx'
+
+local SampledDataset = torch.class("SampledDataset")
```
jungikim: With changes made in e276661, the batch prep logic diverged more and I don't think there's a room for sharing.

jungikim:
> Did you test for efficiency? I measured a 20% slowdown when using -sample_w_ppl with the default options. I guess that comes from the criterion forward step. Not a big deal but this should be taken into account as this dev aims to make the overall training faster.

In my 1mil train, ppl-based sampling is ~10% slower than random sampling due to, as you mentioned, criterion forward step. As stated in the PR request, ppl-based sampling is experimental and the speed up is expected in the reduced number of iterations or from enabling the model better handle more difficult training instances.

onmt/train/Trainer.lua
```diff
@@ -254,10 +271,24 @@ function Trainer:train(model, optim, trainData, validData, dataset, info)
       end
     end
 
+    -- if needed, get avgPpl from epochState before log() since log() will reset internal stats
+    local avgPpl
+    if self.options.sample and self.options.sample_w_ppl then
+      avgPpl = epochState:getAvgPpl()
```
jungikim: Reverted with 8c3fcf3. I didn't know epochState keeps only the stat from last batch, The module name implies otherwise.

onmt/modules/Decoder.lua
jungikim: fixed in 9630b83

onmt/modules/Decoder.lua
jungikim: fixed in 9630b83

onmt/modules/Decoder.lua
```diff
@@ -41,6 +41,7 @@ function Decoder:__init(inputNetwork, rnn, generator, inputFeed)
   -- vector each time representing the attention at the
   -- previous step.
   self.args.inputFeed = inputFeed
+  self.args.indvLoss = indvLoss
```
jungikim: Good idea not adding it in serialized attribute. Fixed in 4c70ca2.

jungikim:
@guillaumekln Updated branch passes luacheck, test/test.lua, and runs on sample data with/without case feature.
I think this is ready for another review.

train.lua
```diff
@@ -11,6 +11,13 @@ local modelClass = onmt.ModelSelector(modelType)
 local options = {
   {'-data',       '', [[Path to the training *-train.t7 file from preprocess.lua]],
                       {valid=onmt.utils.ExtendedCmdLine.nonEmpty}},
+
+  {'-sample',              0, [[Number of instances to sample from train data in each epoch]]},
+  {'-sample_w_ppl',    false, [[use ppl as probability distribution when sampling]]},
+  {'-sample_w_ppl_init',  15, [[start perplexity-based sampling when average train perplexity per batch falls below this value]]},
+  {'-sample_w_ppl_max', -1.5, [[when greater than 0, instances with perplexity above this value will be considered as noise and ignored;
+                                when less than 0, mode + (-sample_w_ppl_max) * stdev will be used as threshold]]},
+
```
jsenellart-systran: these options should be moved, probably to SampleDataSet

onmt/data/SampledDataset.lua
```diff
@@ -0,0 +1,266 @@
+--[[ Data management and batch creation. Handles data created by `preprocess.lua`. ]]
+require 'torchx'
```
jsenellart-systran: considering the size of AliasMultinomial.lua - we should rather integrate it in onmt as we used to integrate LinearNoBias - waiting for the module to pass into standard libraries

onmt/data/SampledDataset.lua
```diff
@@ -0,0 +1,224 @@
...
+
+local SampledDataset = torch.class("SampledDataset")
+
+--[[ Initialize a data object given aligned tables of IntTensors `srcData`
+  and `tgtData`.
+--]]
+function SampledDataset:__init(srcData, tgtData, samplingSize, sample_w_ppl, sample_w_ppl_init, sample_w_ppl_max)
+
+  self.src = srcData.words
```
jsenellart-systran: @Jungi - in #142, I introduce the ability to replace word input by direct feature input (srcData.vectors) - here your code fully assume that it is words, can you anticipate this change and make your code more generic?

onmt/Seq2Seq.lua
```diff
@@ -73,6 +73,16 @@ function Seq2Seq.dataType()
   return 'bitext'
 end
 
+function Seq2Seq:returnIndividualLosses(enable)
+  if torch.getmetatable(torch.type(self.models.decoder))['returnIndividualLosses'] == nil then
```
guillaumekln: This could be simplified to:

```lua
if not self.models.decoder.returnIndividualLosses then
```

train.lua
```diff
@@ -107,6 +128,13 @@ local function main()
     end
   end)
 
+  if opt.sample_w_ppl then
+    if torch.getmetatable(torch.type(model))['returnIndividualLosses'] == nil or model:returnIndividualLosses(true) == false then
```
guillaumekln: Same simplification as above:

```lua
if not model.returnIndividualLosses
```

onmt/data/SampledDataset.lua
guillaumekln: OK.

onmt/data/SampledDataset.lua
guillaumekln: OK.

onmt/data/SampledDataset.lua
guillaumekln: OK. Let's put it in the `onmt/data` directory then with the license in the file header.

onmt/Seq2Seq.lua
jungikim: Fixed in 08ad2ed

train.lua
jungikim: Fixed in 08ad2ed

train.lua
jungikim: I moved the options to SampledDataset as suggested (08ad2ed)

onmt/data/SampledDataset.lua
jungikim: Retrieving main feature from srcData and tgtData is now changed to "srcData.words or srcData.vectors" and  "tgtData.words or tgtData.vectors"; I hope this is enough. (08ad2ed)

onmt/data/SampledDataset.lua
jungikim: Thank you.

onmt/data/SampledDataset.lua
jungikim: I reverted back to preparing batch by iterating over all instances due to batch efficiency and to keep correct attention weights without using maskedsoftmax.

However, to allow oversampling, Instead of adding one instance when it is marked as sampled, now instances can be added multiple time.

This will potentially bias the batch towards more frequently sampled ones and will have more effect (whether positive or negative) on training than before.

onmt/data/SampledDataset.lua
jungikim: Added: 186a2ad

guillaumekln:
You should update the tests also.

jungikim:
I fixed test (6d5da5e)

PR #155, opened 2017-03-02
Sequence tagger

jsenellart-systran:
Introduce sequence tagger using encoder/generator.

* trains simply with `-model_type seqtagger` on bitext datatype
* supports source/target features - for target, a new option `-time_shift_feature 0` disable time_shifting introduced for seq2seq
* apply model with `tag.lua`

Tested on PennTreeBank WSJ corpus and achieves 96.7% with default parameters, on WSJ 22.

PR also fixes Language Model loading and adds check on `translate.lua` model type.


onmt/translate/Translator.lua
```diff
@@ -28,11 +28,15 @@ end
 
 function Translator:__init(args)
   self.opt = args
-  onmt.utils.Cuda.init(self.opt)
```
guillaumekln: Translation servers expect the `Translator` to initialize the GPU. We should either keep this line or update the servers code accordingly.

onmt/tagger/Tagger.lua
```diff
@@ -0,0 +1,177 @@
...
+end
+
+
+function Tagger:__init(args)
+  self.opt = args
+
+  _G.logger:info('Loading \'' .. self.opt.model .. '\'...')
+  self.checkpoint = torch.load(self.opt.model)
+
+  if self.checkpoint.options.model_type and self.checkpoint.options.model_type ~= 'seqtagger' then
```
guillaumekln: The condition and the error message are wrong.

```lua
if not self.checkpoint.options.model_type or self.checkpoint.options.model_type ~= 'seqtagger' then
```

onmt/translate/Translator.lua
jsenellart-systran: I propose you update the server code - because we have a strange mix here. what do you think?

rocks/opennmt-scm-1.rockspec
```diff
@@ -70,6 +70,8 @@ build = {
     ["onmt.translate.BeamSearcher"] = "onmt/translate/BeamSearcher.lua",
     ["onmt.translate.Translator"] = "onmt/translate/Translator.lua",
     ["onmt.translate.PhraseTable"] = "onmt/translate/PhraseTable.lua",
+    ["onmt.tagger.init"] = "onmt/tagger/init.lua",
+    ["onmt.tagger.Tagger"] = "onmt/tagger/Tagger.lua",
```
jsenellart-systran: done in 672fea0

PR #157, opened 2017-03-05
Complex encoder

jsenellart-systran:
refactoring of Complex Encoder - implements dbrnn, pdbrnn. tested in training and translation.


onmt/modules/BiEncoder.lua
```diff
@@ -42,27 +59,39 @@ Parameters:
...
-  self.args.numEffectiveLayers = rnn.numEffectiveLayers
+  -- Compute rnn hidden size depending on hidden states merge action.
+  if self.args.brnn_merge == 'concat' then
+    if args.rnn_size % 2 ~= 0 then
+      error('in concat mode, rnn_size must be divisible by 2')
+    end
+    args.rnn_size = args.rnn_size / 2
+  end
+
+  self.args.rnn_size = args.rnn_size
```
guillaumekln: As said in another PR, renaming:

* `self.args.rnnSize` to `self.args.rnn_size`
* and `self.args.merge` to `self.args.brnn_merge`

is actually breaking existing `BiEncoder` models as the `self.args` table is serialized and used in the forward and backward methods. A workaround would be to add:

```lua
self.args.rnn_size = self.args.rnn_size or self.args.rnnSize
self.args.brnn_merge = self.args.brnn_merge or self.args.merge
```

in the `load` function.

onmt/modules/GRU.lua
```diff
@@ -28,7 +28,7 @@ Parameters:
   * `layers` - Number of layers
   * `inputSize` - Size of input layer
   * `hiddenSize` - Size of the hidden layers
-  * `dropout` - Dropout rate to use (should be in $$[0,1]$$ range.
+  * `dropout` - Dropout rate to use (in $$[0,1]$$ range, if negative, force dropout even on first layer).
```
guillaumekln: Maybe a flag would be clearer than a negative value? Like `-dropout_input`?

onmt/data/Batch.lua
```diff
@@ -75,6 +75,8 @@ function Batch:__init(src, srcFeatures, tgt, tgtFeatures)
 
   self.sourceInputFeatures = {}
   self.sourceInputRevFeatures = {}
+  -- will be used to return extra padded value
+  self.padTensor = torch.LongTensor(self.size):fill(onmt.Constants.PAD)
```
guillaumekln: What is the use of just returning padding?

onmt/data/BatchTensor.lua
```diff
@@ -0,0 +1,17 @@
+local BatchTensor = torch.class('BatchTensor')
```
guillaumekln: This file should also be added in the rockspec.

onmt/modules/BiEncoder.lua
jsenellart-systran: done in 37b96fe

onmt/modules/GRU.lua
jsenellart-systran: not sure we want to do that since it won't be used by end-user. We are just stacking Encoders in complex encoders - and only in these cases, we will want to have first layers adding a dropout

onmt/data/Batch.lua
jsenellart-systran: in pyramidal encoders - when calculating timestep t for level N, we will need to check value of time k*t, k*t+1, k*t+n. at the border, we will have sometime some overflow that is just handled with this automatic-padding

onmt/data/BatchTensor.lua
jsenellart-systran: done in  75b512b

jsenellart-systran:
for me they are tested enough to pass (and used in https://github.com/jsenellart-systran/OpenNMT/tree/asr)

onmt/modules/GRU.lua
jsenellart-systran: good idea - I added the option in 151a48a. the value is internal set when assembling complex encoders, but we can possibly use it for adding dropout between lookup and first LSTM

PR #158, opened 2017-03-06
Add partial CuDNN support

guillaumekln:
This PR adds a cuDNN-powered encoder that supports unidirectional LSTM and GRU. Conversion to `nn` is also provided for model release and translation.

### DIFF
--- a/onmt/modules/CudnnEncoder.lua
+++ b/onmt/modules/CudnnEncoder.lua
@@ -74,6 +74,22 @@ function CudnnEncoder:resetPreallocation()
   self.rnn.gradCellOutput = torch.Tensor()
 end
 
+function CudnnEncoder:toNN()
+  local rnn
+
+  if self.rnn.mode == 'CUDNN_GRU' then
+    rnn = onmt.GRU(self.rnn.numLayers, self.rnn.inputSize, self.rnn.hiddenSize, self.rnn.dropout)
+  elseif self.rnn.mode == 'CUDNN_LSTM' then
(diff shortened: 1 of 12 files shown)


onmt/modules/CudnnEncoder.lua
```diff
@@ -0,0 +1,134 @@
...
+]]
+function CudnnEncoder:__init(inputNetwork, rnnClass, layers, inputSize, rnnSize, dropout)
+  parent.__init(self)
+
+  self.rnn = rnnClass(inputSize, rnnSize, layers, false, dropout, false)
+  self.inputNetwork = inputNetwork
+
+  self.net = nn.Sequential()
+    :add(self.inputNetwork)
+    :add(nn.Transpose({1, 2})) -- By default, batch is second in CuDNN.
```
guillaumekln: You are right, it is cleaner. However, I don't think it will be more efficient as cuDNN is doing the transposition internally: https://github.com/soumith/cudnn.torch/blob/master/RNN.lua#L254-L256

anoidgit: Yes, so I say "maybe" with "a little bit", "a little bit" only because `nn.Sequential` need to handle one less module. @guillaumekln

guillaumekln:
Closing this one for now. The gain was not significant and it is another encoder type to maintain.

I make the branch available in the main repository for reference and maybe future integration:

https://github.com/OpenNMT/OpenNMT/tree/cudnn

PR #162, opened 2017-03-13
Fix inconsistent dict file name.

anoidgit:
If you want to load a feature dict file with `-features_vocabs_prefix`, it will find `*.train_feature_1.dict`, but actually it is `*.source_feature_1.dict`.


guillaumekln:
Good catch, thanks!

anoidgit:
My pleasure. @guillaumekln

PR #166, opened 2017-03-15
Support padded source batches during training

jsenellart-systran:
new options - useful for very long sequences (summarization, asr) or potentially for translation.

In preprocess, possibility not to sort sentences by length
* `-sort`: If 1, sort the sentences by size. [1]

In train, possibility to built up batches with uneven sentence length
* `-uneven_batches`: If true, batches are filled up to max_batch_size even if they have different sizes. [false]

### DIFF
--- a/onmt/Seq2Seq.lua
+++ b/onmt/Seq2Seq.lua
@@ -88,8 +88,8 @@ function Seq2Seq:getOutput(batch)
 end
 
 function Seq2Seq:maskPadding(batch)
-  self.models.encoder:maskPadding()
   if self.args.uneven_batches then
+    self.models.encoder:maskPadding()
     if batch.uneven then
       self.models.decoder:maskPadding(batch.sourceSize, batch.sourceLength)
     else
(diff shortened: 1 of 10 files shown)


vince62s:
what happens if sort = 0 (no sorting) but -uneven_batches remains to false (no fill up) ?
we end up with potentially very small (size=1) batches ?

jsenellart-systran:
yes

onmt/Seq2Seq.lua
```diff
@@ -85,12 +87,32 @@ function Seq2Seq:getOutput(batch)
...
+end
+
 function Seq2Seq:forwardComputeLoss(batch)
+  self:maskPadding(batch)
   local encoderStates, context = self.models.encoder:forward(batch)
   return self.models.decoder:computeLoss(batch, encoderStates, context, self.criterion)
 end
 
 function Seq2Seq:trainNetwork(batch, dryRun)
+  if self.args.uneven_batches then
```
guillaumekln: Let just call `self:maskPadding(batch)` to avoid duplication even if `self.models.encoder:maskPadding()` does nothing on a second call.

onmt/Seq2Seq.lua
```diff
@@ -85,12 +87,32 @@ function Seq2Seq:getOutput(batch)
   return batch.targetOutput
 end
 
+function Seq2Seq:maskPadding(batch)
+  self.models.encoder:maskPadding()
```
guillaumekln: Shouldn't this line also be  encapsulated by `if self.args.uneven_batches then`?

onmt/Seq2Seq.lua
jsenellart-systran: yes - fixed in 700f74e

onmt/Seq2Seq.lua
jsenellart-systran: yes - fixed in 700f74e

PR #167, opened 2017-03-17
support batch for rest_server

vince62s:
This is for batch support in the rest_server.
Hopefully will make things quicker.

### DIFF
--- a/tools/rest_translation_server.lua
+++ b/tools/rest_translation_server.lua
@@ -42,20 +42,23 @@ cmd:option('-batchsize', 1000, [[Size of each parallel batch - you should not ch
 
 local opt = cmd:parse(arg)
 
-local function translateMessage(translator, req)
+local function translateMessage(translator, lines)
   local batch = {}
   -- We need to tokenize the input line before translation
-  local srcTokens = {}
+
   local bpe
-  local srcTokenized = {}
-  local res
-  local err
-  local tokens
+
   _G.logger:info("Start Tokenization")
   if opt.bpe_model ~= '' then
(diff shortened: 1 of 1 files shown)


tools/rest_translation_server.lua
```diff
@@ -124,9 +126,11 @@ local function init_server(port, translator)
       consumes = "application/json",
       produces = "application/json",
       handler = function(req)
-        _G.logger:info("receiving request: [%s]", req.src:gsub("\n", "\\n"))
+--        print(req)
```
guillaumekln: Could you just remove this commented line for good measure?

tools/rest_translation_server.lua
```diff
@@ -124,9 +126,11 @@ local function init_server(port, translator)
...
+        _G.logger:info("receiving request")
         local translate = translateMessage(translator, req)
         _G.logger:info("sending response")
+--        print(translate)
```
guillaumekln: Here also.

tools/rest_translation_server.lua
```diff
@@ -64,53 +65,54 @@ local function translateMessage(translator, req)
...
-    }
-    if opt.withAttn or req.withAttn then
-      local attnTable = {}
-      for j = 1, #results[1].preds[i].attention do
-        table.insert(attnTable, results[1].preds[i].attention[j]:totable())
+      local lineres = {
+        tgt = oline,
+        src = srcSent,
+        n_best = i,
+        pred_score = results[1].preds[i].score
```
guillaumekln: I think this should be `results[b]`.

PR #171, opened 2017-03-22
Dynamic parameter changes

jsenellart-systran:
* more controls on parameters when reloading model
* make difference between default parameters and user-set parameters
* implement dropout dynamic change

### DIFF
--- a/onmt/train/Checkpoint.lua
+++ b/onmt/train/Checkpoint.lua
@@ -127,7 +127,7 @@ function Checkpoint.loadFromCheckpoint(opt)
     for k,v in pairs(structural_parameters) do
       -- if parameter was set in commandline (and not default value)
       -- we need to check that we can actually change it
-      if opt[k] and not opt[k..'_default'] and opt[k] ~= checkpoint.options[k] then
+      if opt[k] and not (opt._is_default and opt._is_default[k]) and opt[k] ~= checkpoint.options[k] then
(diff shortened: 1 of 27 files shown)


onmt/utils/ExtendedCmdLine.lua
```diff
@@ -209,15 +215,23 @@ function ExtendedCmdLine:dumpConfig(opt, filename)
...
 
 function ExtendedCmdLine:parse(arg)
   local i = 1
-  local params = self:default()
+
+  -- set default value
+  local params = {}
+  for option,v in pairs(self.options) do
+    params[strip(option)] = v.default
+    params[strip(option)..'_default'] = true
```
guillaumekln: It seems simpler and clearer to use a sub-table `params.is_default[strip(option)]` rather than building the key every time.

onmt/train/Checkpoint.lua
```diff
@@ -76,23 +76,81 @@ function Checkpoint:saveEpoch(validPpl, epochState, verbose)
   self:save(filePath, info)
 end
 
+-- structural parameters - the parameters with true can be changed dynamically in a train_from
+local structural_parameters = {
```
jsenellart-systran: yes I agree -but the current implementation was even less future proof :smile: with hard coded set-up of options. I think we do need 3 values - I will take the point.

PR #176, opened 2017-03-28
Add n-best list and embedding output options

asaluja:
To: @guillaumekln 

What: I added the ability to print out the n-best list when using the `-n_best` flag, and also print out the embeddings of the input source text using the `-print_input_embeddings` flag.

### DIFF
--- a/onmt/translate/Translator.lua
+++ b/onmt/translate/Translator.lua
@@ -67,11 +67,11 @@ local options = {
       If set to 0, no EOS normalization.]]
   },
   {
-    'dump_input_encoding', false,
+    '-dump_input_encoding', false,
     [[Instead of generating target tokens conditional on
     the source tokens, we print the representation
     (encoding/embedding) of the input.]]
-  }    
+  }
 }
 
 function Translator.declareOpts(cmd)
--- a/onmt/translate/Translator.lua
+++ b/onmt/translate/Translator.lua
@@ -18,7 +18,9 @@ local options = {
   {'-pre_filter_factor', 1, [[Optional, set this only if filter is being used. Before
                             applying filters, hypotheses with top `beamSize * preFilterFactor`
                             scores will be considered. If the returned hypotheses voilate filters,
-                            then set this to a larger value to consider more.]]}
+                            then set this to a larger value to consider more.]]},
+  {'-print_input_embeddings', false, [[Instead of generating target tokens conditional on
+                            the source tokens, we print the representation (embedding) of the input.]]}
 }
 
 function Translator.declareOpts(cmd)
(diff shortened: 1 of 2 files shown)


translate.lua
```diff
@@ -32,7 +32,6 @@ local function main()
   local opt = cmd:parse(arg)
 
   _G.logger = onmt.utils.Logger.new(opt.log_file, opt.disable_logs, opt.log_level)
-  onmt.utils.Cuda.init(opt)
```
guillaumekln: Why was this line removed?

translate.lua
```diff
@@ -107,30 +110,35 @@ local function main()
...
+                  _G.logger:info('BEST HYP:')
+                end
+              end
 
               if #results[b].preds > 1 then
-                _G.logger:info('')
-                _G.logger:info('BEST HYP:')
+                _G.logger:info("[%.2f] %s", results[b].preds[n].score, sentence)
+                if n ~= 1 then -- for the n == 1 case, we already wrote out the hypothesis above
+                  outFile:write(sentence .. '\n')
```
guillaumekln: There is no need to have special cases for `n == 1` and `n ~= 1`. Just call `write` at the beginning of the loop for all `n`.

onmt/translate/Translator.lua
```diff
@@ -18,7 +18,9 @@ local options = {
   {'-pre_filter_factor', 1, [[Optional, set this only if filter is being used. Before
                             applying filters, hypotheses with top `beamSize * preFilterFactor`
                             scores will be considered. If the returned hypotheses voilate filters,
-                            then set this to a larger value to consider more.]]}
+                            then set this to a larger value to consider more.]]},
+  {'-print_input_embeddings', false, [[Instead of generating target tokens conditional on
```
guillaumekln: I find this option name ambiguous mostly because embeddings usually refer to word embeddings. What about `dump_input_encoding`?

onmt/translate/Translator.lua
```diff
@@ -167,84 +169,87 @@ function Translator:translateBatch(batch)
...
   local encStates, context = self.models.encoder:forward(batch)
-
-  -- Compute gold score.
-  local goldScore
-  if batch.targetInput ~= nil then
-    if batch.uneven then
-      self.models.decoder:maskPadding(batch.sourceSize, batch.sourceLength)
+  if self.opt.print_input_embeddings then
+    return encStates
+  else
```
guillaumekln: Could you end the `if` block right there and keep the indentation of the rest of the function? Just so your PR does not conflict with others.

onmt/translate/Translator.lua
```diff
@@ -167,84 +169,87 @@ function Translator:translateBatch(batch)
...
 
   local encStates, context = self.models.encoder:forward(batch)
-
-  -- Compute gold score.
-  local goldScore
-  if batch.targetInput ~= nil then
-    if batch.uneven then
-      self.models.decoder:maskPadding(batch.sourceSize, batch.sourceLength)
+  if self.opt.print_input_embeddings then
+    return encStates
```
guillaumekln: `encStates` contains the final cell and hidden states of all layers. Don't you want to just return the last hidden state `encStates[#encStates]`?

onmt/translate/Translator.lua
```diff
@@ -274,18 +279,32 @@ function Translator:translate(src, gold)
...
-
-      results[b].preds = {}
-      for n = 1, self.opt.n_best do
-        results[b].preds[n] = {}
-        results[b].preds[n].words = self:buildTargetWords(pred[b][n], src[indexMap[b]].words, attn[b][n])
-        results[b].preds[n].features = self:buildTargetFeatures(predFeats[b][n])
-        results[b].preds[n].attention = attn[b][n]
-        results[b].preds[n].score = predScore[b][n]
+      if self.opt.print_input_embeddings then
+        results[b] = encStates
```
guillaumekln: Each batch should have its own encoding. If `translateBatch` returns the last hidden state, it's just about doing:

```lua
results[b] = encStates[b]
```

asaluja:
@guillaumekln thanks so much for your comments! I will address your comments. It also seems like changes to `Translator.lua` have caused some conflicts, so I will also rebase.  I'll try to get to it as soon as I can.

asaluja:
@guillaumekln please take a look.  I've made the changes you requested (thanks for providing those suggestions, I'm not too familiar with the codebase and lua so that was helpful) and resolved merge conflicts. 

One quick question: `encStates[#encStates]` should return the final hidden state in the last layer, right?

guillaumekln:
Looks good to me.

Yes, `encStates[#encStates]` is the final hidden state in the last layer.

PR #178, opened 2017-03-31
Save and Restore RNG States

jsenellart-systran:
when saving checkpoint - save also RNG States so that `train_from` starts from exactly the same state

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * Fix individual losses that were always computed when using random sampling
 * Allow to change the state of word embeddings optimization during a retraining
 * Check consistency of option settings when training from existing model
+* Save and Restore random number generator states in checkpoints
 
 ## [v0.5.3](https://github.com/OpenNMT/OpenNMT/releases/tag/v0.5.3) (2017-03-30)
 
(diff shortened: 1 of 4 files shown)


vince62s:
@guillaumekln @jsenellart-systran ,
For my need ( I use the train_from without the -continue option) I removed the 2 lines here:
https://github.com/jsenellart-systran/OpenNMT/blob/70b7749952c06cfb09c48c5791d340a422484d54/onmt/train/Checkpoint.lua#L136-L137
I think they are not mandatory. Can yo please confirm ?

Otherwise, we can just retrieve the RG states info just before setting to NIL and restate it.

Also, I really think that this "-continue" option is a little confusing.
What about:
if we train_from then behavior is the same as -continue EXCEPT for all options that are passed through command line ?
To me it would be clearer.

guillaumekln:
If `checkpoint.info` is not `nil` the training will just reuse the batch order (for intermediate models) and optimization states (for non SGD methods) of the previous model.

---

For the `-continue` behavior, maybe it is worth opening a thread on the forum to discuss about it. With @jsenellart-systran recent changes we are now able to know whether an option was set or not (it was not the case before) so this new approach can be considered.

vince62s:
But this is the same behavior as with -continue, no ?

guillaumekln:
Yes, that is why previous training information are cleared when not using `-continue`.

PR #190, opened 2017-04-06
Global attention

jsenellart-systran:
possibility to disable attention using `-attention none` and introduce the 3 types of global attention from Luong 2015


onmt/utils/MemoryOptimizer.lua
```diff
@@ -13,6 +13,7 @@ local protectOutput = {
 local protectInput = {
   'nn.Linear',
   'nn.JoinTable',
+  'JoinReplicateTable',
```
guillaumekln: I think this should be `onmt.JoinReplicateTable`.

onmt/modules/JoinReplicateTable.lua
```diff
@@ -0,0 +1,140 @@
+local JoinReplicateTable, parent = torch.class('onmt.JoinReplicateTable', 'nn.Module')
```
guillaumekln: Does this file come from somewhere else? To be consistent with the rest of the project, indentation should be 2 spaces and variable names should be in camelCase.

onmt/modules/Decoder.lua
```diff
@@ -226,7 +234,7 @@ Returns:
  1. `out` - Top-layer hidden state.
  2. `states` - All states.
 --]]
-function Decoder:forwardOne(input, prevStates, context, prevOut, t)
+function Decoder:forwardOne(input, _, prevStates, context, prevOut, t)
```
guillaumekln: I know it is for preparing the coverage attention, but ideally this change should not be part of this PR. Same for `initializeSpecialStates` and `cloneId`.

jsenellart-systran: ok (tu es dur en affaire 😬 )!

onmt/modules/JoinReplicateTable.lua
jsenellart-systran: base of the file was taken from nn.JoinTable - indentation fixed in 74393f0

onmt/Factory.lua
```diff
@@ -244,6 +253,16 @@ function Factory.buildGenerator(rnnSize, dicts)
   end
 end
 
+function Factory.buildAttention(args)
```
jsenellart-systran: seems a good practice to have each module receiving the options?

guillaumekln: OK, we are indeed doing it for encoders.

However, I still think the attention module should actually be built here (otherwise the function name is ambiguous as just the class is returned).

PR #196, opened 2017-04-10
WIP: Multi model Rest server

vince62s:
This is a preliminary, yet working, version of a multi model Rest server.
Breaking changes:
Now including source code of restserverlua and xavante.lua as part of the code.

Unique IP/port listening, but now request need to include "id" : n, where n is the model number in the list.

Need to outsource the models list (in a YAML file) instead of part of the code.

Principle is as follow:
Load model as needed, and unload after 15 seconds of idle time.
TODO: check available memory and wait before loading additional model (to avoid OOM)

Before going further, review is needed. @guillaumekln  ?


guillaumekln:
We should put the whole license in external files header. See for example https://github.com/OpenNMT/OpenNMT/blob/master/onmt/data/AliasMultinomial.lua.

vince62s:
done.
@guillaumekln : here https://github.com/OpenNMT/OpenNMT/pull/196/files#diff-56d1c0c69bc0ce8f30dc5e71d1a67ba9R165
I have a simple solution which is to check if available memory is more than X MB (const to set at the top) before loading the model, or if there is a magic formula to get an estimate of the model size in Ram (based on the filesize) then it might be used.

jsenellart-systran:
Vincent - I added the support for yaml configuration - can you check? the yaml file should contain something like that:

```
-
  model: '/nmt/NMTModels/fr-en/model-fren_epoch10_4.10.t7'
  replace_unk: true
-
  model: '/nmt/NMTModels/en-fr/model-enfr_epoch10_3.42.t7'
  replace_unk: true
```

`-port` is a server option, and I moved also `-withAttn` at the server level although it can also be defined for a given model.
Unload time is defined by `-unload_time`.

tools/rest_multi_models.lua
```diff
@@ -0,0 +1,250 @@
...
+fconfig:close()
+
+local server_cfg = yaml.load(content_config)
+
+-- CAREFULL opt is a table of options here, ie a table of what opt is usually in onmt
+
+local opt = {}
+
+for i=1, #server_cfg do
+  local model_cmd = onmt.utils.ExtendedCmdLine.new('rest_translation_server.lua')
```
vince62s: I think there is a mistake which was in the original code.
it should be 'rest_mulit_models.lua'

tools/rest_multi_models.lua
```diff
@@ -0,0 +1,250 @@
...
+  onmt.translate.Translator.declareOpts(model_cmd)
+  onmt.utils.Cuda.declareOpts(model_cmd)
+  onmt.utils.Logger.declareOpts(model_cmd)
+  tokenizer.declareOpts(model_cmd)
+
+  model_cmd:text("")
+  model_cmd:text("**Other options**")
+  model_cmd:text("")
+  model_cmd:option('-batchsize', 1000, [[Size of each parallel batch - you should not change except if low memory.]])
+  opt[i] = model_cmd:parse({})
```
vince62s: It no longer takes the "-gpuid 1" option as is. I can't figure out why exactly.

tools/rest_multi_models.lua
jsenellart-systran: actually, it is not relevant here since `model_cmd` is just simulating translation server cmdline

tools/rest_multi_models.lua
jsenellart-systran: added the option at server level in 8eac334

tools/rest_multi_models.lua
```diff
@@ -0,0 +1,251 @@
...
+  -- Translate
+  _G.logger:info("Start Translation")
+  local results = translator:translate(batch)
+  _G.logger:info("End Translation")
+
+  -- Return the nbest translations for each in the batch.
+  local translations = {}
+  for b = 1, #lines do
+    local ret = {}
+    for i = 1, translator.opt.n_best do
```
vince62s: we need to change this .opt. by .args. It was changed in a previous commit in master. I had it on my local changes but never pushed it.

vince62s:
Now that it's cleaner, we may add a check on model file existence during the yml loading process.

jsenellart:
Hi Vincent, it seems that it is more than time to merge this PR. could you please add a bit in documentation http://opennmt.net/OpenNMT/tools/servers/#rest ? thx.

jsenellart:
thanks Vincent!

PR #226, opened 2017-04-24
Add a bridge layer between the encoder and the decoder

guillaumekln:
This PR adds an extra layer between the encoder and decoder to define how the encoder states are passed (or not) to the decoder. This also allows the encoder and decoder to have a different number of layers.

### DIFF
--- a/onmt/Model.lua
+++ b/onmt/Model.lua
@@ -80,15 +80,17 @@ function Model:initParams()
   local numParams = 0
 
   for i, key in ipairs(orderedIndex) do
-    params[i]:uniform(-self.args.param_init, self.args.param_init)
+    if params[i]:dim() > 0 then
+      params[i]:uniform(-self.args.param_init, self.args.param_init)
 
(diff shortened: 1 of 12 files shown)


onmt/Model.lua
```diff
@@ -80,15 +80,17 @@ function Model:initParams()
   local numParams = 0
 
   for i, key in ipairs(orderedIndex) do
-    params[i]:uniform(-self.args.param_init, self.args.param_init)
+    if params[i]:dim() > 0 then
```
jsenellart-systran: when do you have `params[i]:dim()==0`

guillaumekln: When the `Bridge` layer has no parameters (the default). It goes through this function like the `Encoder` and `Decoder` do.

jsenellart-systran:
the PR looks fine - I think you should just add a stronger consistency check with legacy option: either the `layers` option is used, or the new options - but we should probably not mix them in a fuzzy way.

PR #247, opened 2017-05-01
Importance sampling

jsenellart-systran:
implement target vocabulary importance sampling based on `-sample` option.  When using sample and `-importance_sampling` option, the target vocabulary is adjusted for each epoch to consider in generator for the sample batches.

Inspired from: https://arxiv.org/pdf/1412.2007.pdf

the implementation uses RIndexLinear - with potential gain speed just on the generator/softmax layer - represented here: https://gist.github.com/jsenellart-systran/417247dc955ffa68994dd28be8f80335

the test script is here: https://gist.github.com/jsenellart-systran/9ee8baf61f61d01044d2bf7a22294cc3

first column is output size, second column time of 100 generator iterations (RNNSIZE 1000), third column the cost using sample target vocabulary of 30K.

* time with RIndexLinear does not depend on complete voc size
* we can expect some significant gain starting at voc size ~100k


vince62s:
great! eager to test.
Just wondering why this is "linked" to the -sample option of the dataset.

jsenellart-systran:
this is linked to the sample option - because the mechanism is to restricted the target vocabulary for subset of the training corpus - today, the sample option is exactly preparing subset of the training corpus (I will add an iterative mode that will not select sentences randomly) so it is naturally connected.

onmt/modules/RIndexLinear.lua
```diff
@@ -0,0 +1,42 @@
...
+      self.weight:resize(self.fullWeight:size()):copy(self.fullWeight)
+      self.bias:resize(self.fullBias:size()):copy(self.fullBias)
+    end
+    self.gradWeight:resize(self.weight:size())
+    self.gradBias:resize(self.bias:size())
+    self.rowIndices = rowIndices
+  end
+end
+
+function Linear:RIndex_clean()
```
guillaumekln: If I did not miss something in this PR, model trained with importance sampling can not be used for translation unless released.

Maybe we could override `Linear.write` to restore the full parameters before saving (see for example https://github.com/soumith/cudnn.torch/blob/master/RNN.lua#L851).

onmt/modules/Generator.lua
```diff
@@ -9,28 +10,67 @@ local Generator, parent = torch.class('onmt.Generator', 'onmt.Network')
...
   local generator = nn.ConcatTable()
   local rnn_size = opt.rnn_size
 
   for i = 1, #sizes do
-    local feat_generator
-    feat_generator = nn.Sequential()
-                        :add(nn.Linear(rnn_size, sizes[i]))
-                        :add(nn.LogSoftMax())
-    generator:add(feat_generator)
+    if i == 1 then
```
guillaumekln: Could this be simplified to:

```lua
local expand = nn.Linear(rnn_size, sizes[i])
if i == 1 then
  self.rindexLinear = expand
end
generator:add(nn.Sequential()
  :add(expand)
  :add(nn.LogSoftMax()))
```

onmt/modules/Generator.lua
```diff
@@ -9,28 +10,67 @@ local Generator, parent = torch.class('onmt.Generator', 'onmt.Network')
 -- for back compatibility - still declare FeaturesGenerator - but no need to define it
 torch.class('onmt.FeaturesGenerator', 'onmt.Generator')
 
+local options = {
+  {
+    '-importance_sampling', false,
```
guillaumekln: It seems this option belongs to `SampleDataset`.

test/onmt/ExtendedCmdLineTest.lua
```diff
@@ -100,8 +100,11 @@ end
 function cmdLineTest.fail_unknown()
   local cmd = onmt.utils.ExtendedCmdLine.new()
   onmt.Seq2Seq.declareOpts(cmd)
-
+  onmt.data.SampledDataset.declareOpts(cmd)
+  -- will generate error
+  print('')
```
guillaumekln: This is no more needed as of https://github.com/OpenNMT/OpenNMT/commit/e794053462d6473b35aa8d5d75bbfaf16ce31c79.

onmt/modules/RIndexLinear.lua
jsenellart-systran: it can be used for translation: at the end of each epoch, we call Generator:setTargetVoc() which copy the updated weights and restore the full Linear. The clean is just getting rid off the additional fields `fullWeights` than are useless for inference

onmt/modules/Generator.lua
jsenellart-systran: done

onmt/modules/Generator.lua
jsenellart-systran: for end-user, it is a generator option - the name is not very intuitive since it initially meant "importance sampling of target vocabulary" - while I am not doing sampling any more. an alternative name is welcome

test/onmt/ExtendedCmdLineTest.lua
jsenellart-systran: it is just for conveniency when applying test units - with the print, it shows:
```
146/211 ExtendedCmdLineTest_fail_unknown ................................ [WAIT]        
invalid argument: -xxx
invalid dependency for option -importance_sampling: requires '-sample' option
146/211 ExtendedCmdLineTest_fail_unknown ................................ [PASS]
``` 

otherwise, the first warning is on the [WAIT] line

guillaumekln: Yes but with e794053, `ExtendedCmdLine` no more writes on stderr when used inside the test suite. It throws an actual Lua error which is correctly handled by the `tester:assertError()` method.

onmt/modules/Generator.lua
vince62s: Could it be named "-limit_tgt_vocab" is this just what it is actually doing ?

test/onmt/ExtendedCmdLineTest.lua
jsenellart-systran: indeed - thanks!

onmt/modules/Generator.lua
jsenellart-systran: after long discussions - we opted for "-sample_tgt_vocab", and the option is now part of the sampling options

PR #253, opened 2017-05-02
generic code to pass target to the generator

jsenellart-systran:
- to be used for NCE, importance sampling, or any approximate softmax

### DIFF
--- a/onmt/modules/Decoder.lua
+++ b/onmt/modules/Decoder.lua
@@ -393,12 +393,9 @@ function Decoder:backward(batch, outputs, criterion)
   for t = batch.targetLength, 1, -1 do
     local refOutput = batch:getTargetOutput(t)
 
-    local prepOutputs = outputs[t]
     -- sampling-based generator need outputs during training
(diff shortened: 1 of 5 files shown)


guillaumekln:
This seems over-complicated to me (unless I'm missing something). I would just always pass the reference output and move all the logic inside the generator. So currently it is just about handling `table` inputs:

```lua
function Generator:updateOutput(input)
  input = type(input) == 'table' and input[1] or input
  return parent.updateOutput(self, input)
end

function Generator:updateGradInput(input, gradOutput)
  input = type(input) == 'table' and input[1] or input
  return parent.updateGradInput(self, input, gradOutput)
end
```

Then other generators can use or not the reference output.

I also believe that after calling `evaluate()` (`self.train == false`), the generators should work without reference output (so no need to change anything in `DecoderAdvancer`)

jsenellart-systran:
ok - I was trying to keep the changes minimal or normal workflow but it adds on the readibility. I am passing now systematically target, and generators are supposed to handle tensor or table input.

PR #255, opened 2017-05-04
Layer normalization

jsenellart-systran:


vince62s:
is there a reason why this PR is not merged ? no improvement ?

StevenLOL:
? Any update ?

guillaumekln:
As far as I know, we did not see any gains.

jsenellart:
yes - there is no obvious gain for NMT. I will still prep the request for merge so that it can be tested for other tasks.

PR #295, opened 2017-05-29
Refactor PDBiEncoder and DBiEncoder

guillaumekln:
* Make DBiEncoder a special case of PDBiEncoder with no time reduction
* Add an option to reduce the time dimension with a concatenation instead of a sum as in [Chan et al. 2015](https://arxiv.org/abs/1508.01211)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,13 @@
 ## [Unreleased]
 
+### Breaking changes
+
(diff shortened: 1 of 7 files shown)


vince62s:
I tested it on text. it looks much better. PPL of 4.18 on baseline 1M enfr after 7 epochs. (network of 600 reduction of 2 pdbrnn_merge concat).

guillaumekln:
Thank you for testing this. Very helpful!

PR #297, opened 2017-05-29
[WIP] Function parameter

jsenellart:
add the possibility of dynamically changing an option value depending on epoch, but also potentially on validation ppl delta, current learning rate, etc...

The value of the parameter should a lua expression enclosed in parenthesis `(...)` and can access "state" which is a map of values using `_G.s` namespace. the lua expression type should be compatible with the parameter normal value.

For instance, to switch on dropout with value 0.2 after 6 epochs with value 0, use `-dropout (_G.s.epoch>6 and 0.2 or 0)`

This will also be used for scheduled sampling, and could be used for learning rate tuning, etc..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### New features
 
 * New option in preprocess to check that sizes of source and target are equal (for seqtagging)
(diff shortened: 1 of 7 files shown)


jsenellart:
nice idea but too complicated :)

PR #300, opened 2017-05-31
[WIP] Ensemble decoding

guillaumekln:
Implemented ensemble decoding as requested. Should be ready for use (simple usage described in `docs/translation/ensemble.md`), but I would like to improve the current integration and performance.


vince62s:
Would there be a way, using the distillation technique, to build a single model from an ensemble ?

PR #303, opened 2017-06-01
memory graph

jsenellart-systran:
build full calculation graphs as a first step to improve memory optimization.  accessorily provides visualization functions using `-dump_graphs`


jsenellart-systran:
temporary withdrawing to extend to non-nngraph models

Issue #288, opened 2017-05-23
bug ? using train_from and Importance sampling with 2 GPU


vince62s:
and I get this : (which I don't when using the same command line and 1 GPU only)
```
[05/23/17 15:24:42 INFO] Loading checkpoint 'exp/model-enfr_epoch300_3.74.t7'...	
[05/23/17 15:24:52 INFO] Preparing memory optimization...	
[05/23/17 15:24:53 INFO]  * sharing 71% of output/gradInput tensors memory between clones	
[05/23/17 15:25:15 INFO] Restoring random number generator states...	
[05/23/17 15:25:15 WARNING] GPU count does not match for resetting random number generator. Skipping.	
[05/23/17 15:25:15 INFO] Start training from epoch 301 to 600...	
[05/23/17 15:25:15 INFO] 	
[05/23/17 15:25:15 INFO] Sampling dataset...	
[05/23/17 15:25:17 INFO] Sampled 50000 instances into 1085 batches.	
[05/23/17 15:25:17 INFO] Importance Sampling - keeping 44772 target vocabs.	
/home/torch/install/bin/luajit: ./onmt/train/Trainer.lua:150: attempt to index field 'model' (a nil value)
stack traceback:
	./onmt/train/Trainer.lua:150: in function 'trainEpoch'
	./onmt/train/Trainer.lua:378: in function 'train'
	train.lua:173: in function 'main'
	train.lua:178: in main chunk
	[C]: in function 'dofile'
	.../torch/install/lib/luarocks/rocks/trepl/scm-1/bin/th:150: in main chunk
	[C]: at 0x00406670
```

vince62s:
my bad maybe we cannot train_from with 2 GPU when initially trained with 1 GPU .....

guillaumekln:
It should work actually. I will take a look.

guillaumekln:
It was actually the combination of importance sampling and multi GPUs that did not work. Thanks for reporting!

vince62s:
whatever the batch size I take (here 32 which works with 1 GPU) I get this error : 
```
[05/24/17 07:59:06 INFO] Sampling dataset...	
[05/24/17 07:59:08 INFO] Sampled 50000 instances into 1603 batches.	
[05/24/17 07:59:08 INFO] Importance Sampling - keeping 44772 target vocabs.	
[...]

PR #304, opened 2017-06-01
Optimize memory usage of network replicas

guillaumekln:
Network replicas called `getParameters` even though parameters are already flattened in a single storage.

@vince62s Related to #288, I would like you to test this PR (using `THC_CACHING_ALLOCATOR=0` if needed).

### DIFF
--- a/onmt/Model.lua
+++ b/onmt/Model.lua
@@ -97,7 +97,20 @@ function Model:initParams()
   return params, gradParams
 end
 
-function Model:getParams()
+--[[ Retrieve all parameters and gradients from the models.
+
+Parameters:
+
+* `flattened` - if true, assume that the parameters are already flattened.
+
+Returns:
+
+* all parameters
+* all gradients
+* a table that maps parameters to their model key
+
+]]
+function Model:getParams(flattened)
   -- Order the model table because we need all replicas to have the same order.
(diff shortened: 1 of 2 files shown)


vince62s:
here are some some results:
Without sampling
voc 50k 4x768 vec 500 BS 64
1GPU: memory 4.4GB speed first 500 ite: 2706 tok/sec second 500 it: 2958 tok/sec
2GPU: memory 4.7GB + 5.3 GB first 500 ite: 3039 tok/sec second 500 ite: 3260 tok/sec

With sampling: voc 100k, sampling 50k
1 GPU: memory 4.4GB speed (2785+3134)/2, epoch 2 (3088+3016)/2
2 GPU: memory 4.7GB + 7.1 GB speed epoch1 3038, epoch 2 3376

first comment: speed gain is not huge with 2 GPU
second comment: with IS, see the 2nd GPU memory footprint

I will try with vocab 200k
(also here this is with Lua, not luajit like before)

jsenellart-systran:
@vince62s - could you get the same number without this patch? and if possible for "with sampling", set seq to 50 so that we can compare?
thanks!
speed gain is a different story...

guillaumekln:
Thanks for the experiments @vince62s.

PR #305, opened 2017-06-01
Add case segmenting in order to be able restore words like WiFi

kovalevfm:
With this option tokenizer splits words like WiFi to two words Wi Fi, so we can restore correct casing with detokenizer.
Example:

```
echo "Abc ABC ABc abC AbC" | th tools/tokenize.lua -joiner_annotate -case_feature -mode aggressive -segment_case | th tools/detokenize.lua -case_feature

Abc ABC ABc abC AbC
```

### DIFF
--- a/docs/tools/tokenization.md
+++ b/docs/tools/tokenization.md
@@ -24,6 +24,14 @@ th tools/detokenize.lua OPTIONS < file.tok > file.detok
 * `￨` is the feature separator symbol. If such character is used in source text, it is replaced by its non presentation form `│`.
 * `￭` is the default joiner marker (generated in `-joiner_annotate marker` mode). If such character is used in source text, it is replaced by its non presentation form `■`
 
+## Mixed casing words
+`-segment_case` feature enables tokenizer to segment words in order to restore in the process of  detokenization correct casing for mixed casing words.
+Tokenizer with this option splits words into segments so all letters in every segment are either lowercased, or uppercased, or first letter is uppercases and rest are lowercased
+```text
+WiFi --> Wi￨C Fi￨C
+TVs --> Tv￨C s￨l
+```
+
 ## BPE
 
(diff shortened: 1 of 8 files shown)


guillaumekln:
Thanks for the PR!

I'm wondering if we should make this the default behavior when using `-case_feature`.

@jsenellart-systran What do you think?

jsenellart-systran:
Thanks for the PR! agree it makes 100% sense to combine it with `-case_feature` - but maybe just a warning if it is not switched on? otherwise we will start having an option that will take different default values depending on other options which will be misleading - also the same logic would need to be part of the server mode (if we had thought about that before, it could have been part of aggressive tokenization scheme..).

@kovalevfm - would you mind adding a note on case handling also in the documentation: https://github.com/OpenNMT/OpenNMT/blob/master/docs/tools/tokenization.md?

docs/tools/tokenization.md
```diff
@@ -24,6 +24,13 @@ th tools/detokenize.lua OPTIONS < file.tok > file.detok
 * `￨` is the feature separator symbol. If such character is used in source text, it is replaced by its non presentation form `│`.
 * `￭` is the default joiner marker (generated in `-joiner_annotate marker` mode). If such character is used in source text, it is replaced by its non presentation form `■`
 
+## Mixed casing words
+`-segment_case` feature enables tokenizer to segment words into subwords with one of 3 casing types (truecase ('House'), uppercase ('HOUSE') or lowercase ('house')), which helps  restore right casing during  detokenization. This feature is especially useful for texts with a signficant number of words with mixed casing ('WiFi' -> 'Wi' and 'Fi').
+```text
+WiFi --> Wi￨C Fi￨C
```
guillaumekln: I think this should be:
```text
WiFi --> wi￨C fi￨C
TVs --> tv￨U s￨L
```

kovalevfm: You're right, fixing it.

guillaumekln:
Looks good to me. Thanks!

PR #320, opened 2017-06-19
use nn.Squeeze rather than nn.Sum

anoidgit:


guillaumekln:
Thanks!

anoidgit:
My pleasure @guillaumekln

PR #325, opened 2017-06-21
[WIP] Multi head attention, dot scaled, and attention dropout

jsenellart:
In [Attention Is All You Need](https://arxiv.org/pdf/1706.03762.pdf) paper, several concept are introduced that can fit in our current attention module:

* "Scaled Dot-Product Attention" - improving `dot` model (option `-global_attention dot_scaled`)
* multi-head attention (option `-multi_head_attention N`)
* dropout on attention (option `-dropout_attention`)


vince62s:
Jean , did you get interesting results with the multi head functionality ?

jsenellart:
@vince62s - I cleaned up the implementation - it is simpler now and I get slight but consistent improvements in PPL with 2 heads on 1M dataset with/without dropout_attention (0.1), and general model. On the other hand `dot_scaled` does not seem to improve `dot`.  Could you do some tests on your side too?

vince62s:
First comments:
I did it on a smaller dataset (500k).
If I train with multi_head 2 or 4 from the very begining, PPL explodes.
If I train with Multihead 1 for the first epoch and then 2, I am getting a lower PPL in the end, but not necessarily better BLEU.
If I train with Multihead 1 for the first epoch and then 4, I am getting exactly the same PPL (at 2 decimals) as in the previous experiment (at least for epoch 2 and 3): that does not sound right.

jsenellart:
thanks Vincents. For PPL explosion, I could only get it converge with Adam on my side or by setting lower LR. the problem with SGD and making LR smaller globally is that all the model is penalized. Adam naturally adapts to local variations but on the other hand, we know that we did not get similar results between adam and sgd. maybe by changing after the first epoch?

PR #328, opened 2017-06-27
Add script to average various checkpoints with the same architecture

vince62s:
For review.
I tested it during a sampled training.
I averaged 5 checkpoints from the last 50 epochs.
Very limited improvements.


guillaumekln:
Thanks!

PR #336, opened 2017-07-07
CNN encoder

natsegal:


onmt/modules/CNNEncoder.lua
```diff
@@ -0,0 +1,174 @@
+--[[ Google's Convolutional Encoder as described in https://arxiv.org/abs/1611.02344.
+
```
vince62s: You mean facebook ?

natsegal: Yes, indeed, thanks ! :)

vince62s:
Hi @natsegal , cc @guillaumekln 
I get this error:
[07/07/17 17:27:57 INFO]    - structure: cnn_kernel = 3; cnn_layers = 2; cnn_size = 500;	
[07/07/17 17:27:57 INFO]  * Decoder:	
[07/07/17 17:27:59 INFO]    - word embeddings size: 500	
[07/07/17 17:27:59 INFO]    - attention: global (general)	
[07/07/17 17:27:59 INFO]    - structure: cell = LSTM; layers = 2; rnn_size = 500; dropout = 0.2 (naive)	
[07/07/17 17:27:59 INFO]  * Bridge: copy	
/home/moses/torch/install/bin/luajit: ./onmt/modules/Bridge.lua:26: with the copy bridge type, encoder and decoder must have the same number of layers and hidden size

misusing it ?

natsegal:
Hi @vince62s , you cannot use CNN encoder with the default copy bridge, you need to use either  "none" or "dense". Also, you should probably use it either with adam or with a much smaller learning rate than the usual rnn encoder, I had trouble making it converge otherwise.

vince62s:
great thanks. Did Adam work well for you in some other configs ? I had very hard time to use it with rnn ....
I'll keep you posted.

vince62s:
BTW, paper seems to suggest that POS is mandatory to get competitive results. Do you confirm ?

natsegal:
With CNN encoder adam actually worked better for me than simple SGD. You may also add -dropout_input option for better convergence.
There are still some issues left that need to be resolved, for instance the results with padded batch are still a little worse than with -batch_size 1. I continue working on it.
And yes, positions do help, at least it my settings.

natsegal:
I think we can get competitive results, we just need to find the right configuration for optimization. So far CNN is still 2 BLEU points lower in my settings...

jsenellart:
@natsegal : for your constraint on 'copy' bridge - you can use 'depends' cmdline - to validate that one does not select illegal option combination - see https://github.com/OpenNMT/OpenNMT/blob/master/onmt/data/SampledDataset.lua#L44-L51

PR #338, opened 2017-07-09
Fix LuaJIT out of memory issues in learn_bpe

mousai:
Changed various tables to tds.Vec or tds.Hash to avoid issues with LuaJIT's 2GB heap limit


guillaumekln:
Thanks!

PR #342, opened 2017-07-12
Add option to save the validation translation to a file

guillaumekln:


vince62s:
you write "score ||| output"
I thought we were saying "score ||| output ||| reference"

guillaumekln:
That is correct. I choose not to output the reference for simplicity.

You can still paste the files together, e.g.:

```bash
paste model_epoch1_validation_translation.txt data/tgt-val.txt | sed 's/\t/ ||| /g'
```

vince62s:
not the same order

PR #343, opened 2017-07-13
[WIP] Scheduled sampling

jsenellart:
introduce `scheduled_sampling` for training as defined in [Bengio, 2015].(http://papers.nips.cc/paper/5956-scheduled-sampling-for-sequence-prediction-with-recurrent-neural-networks.pdf)

missing scheduled sampling for features


homink:
Hi Jean, could you extend this to SeqTagger as well?

jsenellart:
Hi Homin, actually it is not relevant with the current implementation of the sequence tagger, we are not using input feeding so state $t+1$ doesn't explicit depend on generated label at state $t$. We could add this as a feature, but I think it is not a common approach for sequence tagging.

PR #362, opened 2017-08-25
Enable Adam learning rate decay and restarts

guillaumekln:
To experiment analysis from [Denkowski et al. 2017](https://arxiv.org/abs/1706.09733).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,14 @@
 ### Breaking changes
 
 * Remove tokenization options from `learn_bpe.lua`
+* Learning rate is also decayed when using Adam
 
 ### New features
 
 * Add TER calculation
 * Add CNN encoder
 * Add option to save the validation translation to a file
+* Add option to reset the optimizer states when the learning rate is decayed
 
 ### Fixes and improvements
 
--- a/docs/options/train.md
+++ b/docs/options/train.md
@@ -89,6 +89,7 @@
 * `-start_decay_at <number>` (default: `9`)<br/>In "default" decay mode, start decay after this epoch.
 * `-start_decay_score_delta <number>` (default: `0`)<br/>Start decay when validation score improvement is lower than this value.
 * `-decay <string>` (accepted: `default`, `epoch_only`, `score_only`; default: `default`)<br/>When to apply learning rate decay. `default`: decay after each epoch past `-start_decay_at` or as soon as the validation score is not improving more than `-start_decay_score_delta`, `epoch_only`: only decay after each epoch past `-start_decay_at`, `score_only`: only decay when validation score is not improving more than `-start_decay_ppl_delta`.
+* `-reset_when_decay [<boolean>]` (default: `false`)<br/>If set, the optimizer states (if any) will be reset when the decay condition is met.
 
 ## Saver options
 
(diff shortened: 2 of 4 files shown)


srush:
@sebastianGehrmann who added this to opennmt-py, we should sync names...

jungikim:
Hi @guillaumekln, I also was testing this feature with my own implementation.
One thing I noticed was that setting each optimState to an empty table resulted in an out of memory error when GPU memory was nearly full, so instead I zeroed the tensors in the tables, for example:

...
  elseif self.args.optim == 'adam' then
    self.optimStates.t = 0
    self.optimStates.m:zero()
    self.optimStates.v:zero()
    self.optimStates.denom:zero()
  end
(I was lazy and didn't implement for the other optim. algorithms ... yet)

vince62s:
I confirm the OOM error just at the epoch it starts to decay

[08/28/17 05:39:41 INFO] Sampling dataset...	
[08/28/17 05:39:42 INFO] Sampled 50000 instances into 822 batches.	
[08/28/17 05:44:53 INFO] Epoch 400 ; Iteration 500/822 ; Optim ADAM LR 0.0002 ; Source tokens/s 2579 ; Perplexity 7.03	
[08/28/17 05:48:12 INFO] Epoch 400 ; Iteration 822/822 ; Optim ADAM LR 0.0002 ; Source tokens/s 2550 ; Perplexity 6.97	
[08/28/17 05:48:12 INFO] Evaluating on the validation dataset...	
[08/28/17 05:48:25 INFO] Validation perplexity: 6.49	
[08/28/17 05:48:25 INFO] 	
[08/28/17 05:48:25 INFO] Sampling dataset...	
[08/28/17 05:48:26 INFO] Sampled 50000 instances into 822 batches.	
THCudaCheck FAIL file=/home/moses/torch/extra/cutorch/lib/THC/generic/THCStorage.cu line=66 error=2 : out of memory
/home/moses/torch/install/bin/luajit: ./onmt/train/Optim.lua:255: cuda runtime error (2) : out of memory at /home/moses/torch/extra/cutorch/lib/THC/generic/THCStorage.cu:66
stack traceback:
	[C]: in function 'new'
	./onmt/train/Optim.lua:255: in function 'adamStep'
	./onmt/train/Optim.lua:128: in function 'prepareGrad'
	./onmt/train/Trainer.lua:258: in function 'trainEpoch'
	./onmt/train/Trainer.lua:425: in function 'train'
	train.lua:173: in function 'main'
	train.lua:178: in main chunk
	[C]: in function 'dofile'
	...oses/torch/install/lib/luarocks/rocks/trepl/scm-1/bin/th:150: in main chunk
	[C]: at 0x00406670

vince62s:
@guillaumekln I got another "issue" to report.
Optim is called from trainer.
When you start to decay at start_decay_at=epoch 400 for instance, it supposed to apply the decayed learning rate for epoch 401.
When it comes to call updatelearningrate at the end of epoch 400, then it will reset the states BEFORE saving the checkpoint, which mean in the end the checkpoint of epoch 400 will take into account the state reset.

I realized this when I saw the size of my checkpoint (1.1GB versus 2.7GB for the previous epochs ...)

I hope my point is clear enough.

guillaumekln:
@srush Ok, I will sync the name with the PyTorch version.

@jungikim Indeed, it seems safer not to rely on the Lua garbage collector. I think we can just loop on the state and zero each `Tensor`. I will add that.

@vince62s Is it unexpected? If that's an issue, it seems you did not want to enable state reset.

vince62s:
well  I would say, what is expected is that the model is saved at epoch 400 with its states, then you reset the states and decay for epoch 401, and model at epoch 401 is saved with the new states.
Anyhow, this i snot working very well. After the state reset, the PPL or whatever metric jumps way too high.

PR #366, opened 2017-09-06
Protected sequences

jsenellart:
tokenization is now protecting sequences inside  ｟ ... ｠- these 2 characters (U+FF5F and U+FF60) are unicode presentation forms so are not supposed to be part of normal text flows.
If there are spaces in the protected block - these spaces are serialized using `％xxxx` where `％` is the unicode presentation form of regular percent `%` and xxxx is unicode sequence. The protection is also protecting from BPE-ization.

@vince62s - can you check it is good for your needs?


francoishernandez:
Hi @jsenellart ,
We ran a first test, here are a few comments:

This apparently doesn't support the case feature for now, as  ｟ ... ｠ has no feature attached, hence the following error:

[09/08/17 17:16:11 INFO] * Building source vocabularies.../home/moses/torch/install/bin/luajit: ./onmt/data/Vocabulary.lua:45: all sentences must have the same numbers of additional features (data/tempo.fr.tok:591921)stack traceback:[C]: in function 'assert'./onmt/data/Vocabulary.lua:45: in function 'make'./onmt/data/Vocabulary.lua:111: in function 'init'tools/build_vocab.lua:52: in function 'main'tools/build_vocab.lua:60: in main chunk[C]: in function 'dofile'...oses/torch/install/lib/luarocks/rocks/trepl/scm-1/bin/th:150: in main chunk[C]: at 0x00406670

Also, having a look at the dictionaries, it seems ｟ ... ｠ followed by other things (without any space) is not split (e.g. ｟ ... ｠st or ｟ ... ｠￭ appear in the dict). It would seem more logical to split such elements, in what @vince62s intended anyways. Yet remains the question of merging it back afterwards...

Thanks!
François

jsenellart:
Thanks François for the feedback - I am taking the feature point. For the split, I had a unit test on that, I will double check. It is supposed not to take the marker so `｟ ... ｠st` => `｟ ... ｠ ￭st` - the only case where it will is when you have a sequence `｟ ... ｠｟ ... ｠`.

jsenellart:
François - I fixed the combination with letters as follow:
```
  testTok(opt, "｟1,023｠km", "｟1,023｠ ￭km", true)
  testTok(opt, "A｟380｠", "A￭ ｟380｠", true)
  testTok(opt, "｟1,023｠｟380｠", "｟1,023｠￭ ｟380｠", true)
  testTok(opt, "｟1023｠.", "｟1023｠ ￭.", true)
```

For the case feature, I do not reproduce the problem - when I use `-case_feature` option in tokenizer, it correctly carries a  `￨N` feature.  Can you double check?

francoishernandez:
Hi Jean,
Thanks for the fix, tested it and it worked perfectly.

I checked the `-case_feature` issue, seems like some old data remained in the pipe and created the error. Everything is working fine now.

francoishernandez:
Hi again,
Seems like there is still one case where ｟...｠ tokens are not isolated properly: when followed directly by a number. E.g. things like "｟...｠0" appear in the dictionaries.
Thanks!
François

francoishernandez:
Also,
Things like "$｟...｠" and "(｟...｠" are not properly tokenized. Some joiner annotates are missing and this leads to outputs with unwanted spaces in decoding, here "$ ｟...｠" and "( ｟...｠".

jsenellart:
I fixed the problem with punctuations before. I can not reproduce a problem with numbers after as shown below. Do you have a sequence?
``` 
  testTok(opt, "$｟0.23｠", "$￭ ｟0.23｠", true)
  testTok(opt, "｟0.23｠$", "｟0.23｠ ￭$", true)
  testTok(opt, "｟US$｠23", "｟US$｠ ￭23", true)
```

vince62s:
I noticed you test script uses -mode conservative.
We use -mode aggressive and also -segment_case
may that matter ?
also under normal conditions $100 would stick the joiner to ￭100 not after the $
is the new output intended to be the reverse ?

francoishernandez:
Jean,
The problem with numbers is seen when the ｟...｠ is placed between two numbers:
"Just over 1￭ ｟...｠0 miles from Guatemala City..."
I will test the behaviour regarding punctuation.

jsenellart:
François - it works for me - can you double check or give me the exact sentence were it fails?
```
  testTok(opt, "1｟ABCD｠0", "1￭ ｟ABCD｠ ￭0", true)
```

Vincent - the rules has always been `letter`>`number`>`other` and the joiner is always attaching to the smaller. Now we add these new object `protected` which is "bigger" that letters.

```
  testTok(opt, "$1", "$￭ 1", true) 
```

francoishernandez:
I checked with your test string and I still have the same behaviour:
`1｟ABCD｠0` --> `1￭ ｟ABCD｠0`
(The options here are: `-mode aggressive -segment_case -joiner_annotate`)

I also added it in the test file and ran it:
`testTok(opt, "1｟ABCD｠0", "1￭ ｟ABCD｠ ￭0", true)`
`Completed 1075 asserts in 254 tests with 1 failure and 0 errors
--------------------------------------------------------------------------------
TokenizerTest_protectedSequence
EQ failed: 1◊￭◊｟ABCD｠◊￭◊0 (string) ~= 1￭◊｟ABCD｠◊￭0 (string)
	test/onmt/TokenizerTest.lua:19: in function 'testTok'
	test/onmt/TokenizerTest.lua:60: in function <test/onmt/TokenizerTest.lua:46>`

Maybe something missing on the repo?

jsenellart:
mmm... I just commited the test unit, let us check what Travis is saying...

francoishernandez:
Pulled again, tests are running fine now.

The issue is with `-mode aggressive`. It is working as intended without the option.
(Tested with tokenize.lua on a file containing the single line `1｟ABCD｠0`.)

jsenellart:
ok - it is supposed to work the same with aggressive mode. I am fixing that and will wrap up this PR. thanks.

jsenellart:
fixed - the issue was pre-existing: `(01.23)` was `( ￭01 ￭. ￭23 ￭)` while it should have been `(￭ 01 ￭.￭ 23 ￭)`

jsenellart:
Thanks @francoishernandez and @vince62s !

PR #370, opened 2017-09-09
Make translation server support the arguments of `onmt.translate.Translator`

anoidgit:


tools/translation_server.lua
```diff
@@ -21,6 +21,7 @@ cmd:setCmdLineOptions(options, 'Server')
 onmt.translate.Translator.declareOpts(cmd)
 onmt.utils.Cuda.declareOpts(cmd)
 onmt.utils.Logger.declareOpts(cmd)
+onmt.translate.Translator.declareOpts(cmd)
```
guillaumekln: This line is already present 2 lines above. Did you encounter issues regarding translation options?

anoidgit: Sorry, maybe my installation of OpenNMT is not up-to-date before. I have encountered some problems(some arguments of `onmt.translate.Translator` like `-coverage_norm` can not work). I have already updated my installation and it is ok now.

PR #372, opened 2017-09-11
-replace_unk_tagged option

bartvanhalder:
Hello Everyone,

I added an option to replace unknown words with the token with the highest attention vector like -replace_unk, but wrapping it in `<unk>Original_token</unk>`. It would make sense to me to not have to maintain a fork just to have this option, so here it is. If there are any problems or suggestions I’m open to changing it.

Regards,
Bart

### DIFF
--- a/onmt/translate/Translator.lua
+++ b/onmt/translate/Translator.lua
@@ -284,7 +284,6 @@ function Translator:buildTargetWords(pred, src, attn)
         elseif self.args.replace_unk_tagged then
           tokens[i] = '｟unk:' .. source .. '｠'
         end
-        
       end
     end
   end
--- a/onmt/translate/Translator.lua
+++ b/onmt/translate/Translator.lua
@@ -42,7 +42,7 @@ local options = {
       does not exist in the table) then it will copy the source token]]},
   {
     '-replace_unk_tagged', false,
-    [[The same as -replace_unk, but wrap the replaced token in <unk>token</unk> if it is not found in the phrase table.]]},
+    [[The same as -replace_unk, but wrap the replaced token in ｟unk:xxxxx｠ if it is not found in the phrase table.]]},
   {
     '-phrase_table', '',
     [[Path to source-target dictionary to replace `<unk>` tokens.]],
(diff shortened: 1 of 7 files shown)


vince62s:
@jsenellart , @guillaumekln 
it is worth discussing this one, because it is linked to both the protected sequences and the do_not_translate option discussed here: http://forum.opennmt.net/t/terminology-handling/750

bartvanhalder:
I'm not completely sure it is linked, a do_not_translate option would act on the input and make sure those tokens do not get translated at all. This pull request adds an option to annotate tokens that were not translated so they can be handled by an external postprocessing step.

vince62s:
you are correct, it's just giving the feeling we are piling up layers to solve the same issue in the end. But I do understand your need.

jsenellart:
still @vince62s is right about the annotation - let us try to keep the same annotation markup used for protected sequences to avoid another type of pseudo annotation parsing - since output is not xml anyway - would you mind switching to a sequence like `｟ unk:xxxxx｠`?

bartvanhalder:
@jsenellart Not at all, is the space between `((` and `unk:xxx))` intentional?

jsenellart:
nope - sorry, I fixed it.

bartvanhalder:
i'm sorry, that last commit does not work because i don't understand lua's handling of utf-8. I should have it fixed in half an hour or so.

onmt/translate/Translator.lua
```diff
@@ -281,6 +284,21 @@ function Translator:buildTargetWords(pred, src, attn)
     end
   end
 
+  if self.args.replace_unk_tagged then
```
guillaumekln: Could you merge this chunk with the loop above? Something like:

```lua
if self.args.replace_unk or self.args.replace_unk_tagged then
  for i = 1, #tokens do
    if tokens[i] == onmt.Constants.UNK_WORD then
      local _, maxIndex = attn[i]:max(1)
      local source = src[maxIndex[1]]

      if self.phraseTable and self.phraseTable:contains(source) then
        tokens[i] = self.phraseTable:lookup(source)
      elseif self.args.replace_unk then
        tokens[i] = source
      elseif self.args.replace_unk_tagged then
        tokens[i] = '｟unk:' .. source .. '｠'
      end
    end
  end
end
```

bartvanhalder: I went for the naive way because I have literally never written more than 3 lines of lua, but I’ll try my hand at more than naively copy pasting and editing :)

bartvanhalder:
I ended up doing exactly what you suggested, i only added a few whitelines for readability

guillaumekln:
Thanks!

bartvanhalder:
Thanks for your time and effort!

PR #385, opened 2017-09-28
Support multi-layer global attention

guillaumekln:


guillaumekln:
This does not bring any improvements. Closing.

PR #386, opened 2017-09-29
Add tokenizer no BPE related options in learn_bpe.lua

DYCSystran:
- Add all no BPE related tokenization options to learn_bpe.lua 's tokenizer with prefix 'tok_'
- Add option '-lowercase' to tokenizer to perform only lowercase but using the same function 'addCase' as the case feature
- Miscelleneous fix for tokenizer's BPE related options' name
- Remove (temporarily) the 'record' mode for BPE models so that the cmd options (-bpe_case_insensitive) would not be overridden by those contained in a BPE model generated by the learn_bpe.lua


tools/utils/tokenizer.lua
```diff
@@ -58,12 +62,12 @@ local options = {
       is learnt using `learn_bpe.lua`.]]
   },
   {
-    '-EOT_marker', separators.EOT,
-    [[Marker used to mark the end of token.]]
+    '-bpe_EOT_marker', separators.EOT,
+    [[Marker used to mark the End of Token while applying BPE in mode 'prefix' or 'both'.]]
```
jsenellart: why changing the name of this option?

DYCSystran: To make it clear as a bpe related option and then for regex based filtering to not include these bpe options in learn_bpe.lua while tokenizing

jsenellart: ok - put a note in *BREAKING CHANGES* section although I don't think it had ever been used. Please prepare also for documentation update and see with Guillaume to merge the PR.

PR #395, opened 2017-10-05
Make decoder embeddings optional in tools/extract_embeddings.lua

pavel-:
When extracting embeddings from SeqTagger or LanguageModel models extract_embeddings throws error since there is no decoder. With this fix we first check if decoder exists, and only then extract embeddings from it.

There was also a typo in SeqTagger.lua description

### DIFF
--- a/tools/extract_embeddings.lua
+++ b/tools/extract_embeddings.lua
@@ -40,8 +40,9 @@ local function main()
   end
   local dicts = checkpoint.dicts
   local encoder = onmt.Factory.loadEncoder(checkpoint.models.encoder)
+  local decoder
   if checkpoint.models.decoder then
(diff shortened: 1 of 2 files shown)


tools/extract_embeddings.lua
```diff
@@ -40,7 +40,9 @@ local function main()
   end
   local dicts = checkpoint.dicts
   local encoder = onmt.Factory.loadEncoder(checkpoint.models.encoder)
-  local decoder = onmt.Factory.loadDecoder(checkpoint.models.decoder)
+  if checkpoint.models.decoder then
+    local decoder = onmt.Factory.loadDecoder(checkpoint.models.decoder)
```
guillaumekln: This variable will not be defined outside of the `if` scope. I think you can declare it just before line 59.

PR #408, opened 2017-10-19
Add keep rules in file sampling to keep all sentences from training files

guillaumekln:
This PR adds a way to keep all sentences from some training files instead of sampling sentences from them. These sentences will always be part of the current sample.

It uses `*` as weight value in the distribution file.

### DIFF
--- a/onmt/data/Preprocessor.lua
+++ b/onmt/data/Preprocessor.lua
@@ -1053,6 +1053,10 @@ function Preprocessor:makeData(dataset, dicts)
       else
         sampledCount = sampledCount - self.keepCount
       end
+      if self.totalCount == 0 and self.keepCount < self.args.gsample then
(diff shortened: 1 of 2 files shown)


onmt/data/Preprocessor.lua
```diff
@@ -1012,18 +1050,29 @@ function Preprocessor:makeData(dataset, dicts)
       local sampledCount = self.args.gsample
       if sampledCount < 1 then
         sampledCount = sampledCount * self.totalCount
+      else
+        sampledCount = sampledCount - self.keepCount
+      end
+      if sampledCount < 0 then
+        _G.logger:error('You requested a sample of %d sentences but %d are already reserved by keep rules. You should configure a larger sample or keep less sentences.',
```
jsenellart: works perfectly but I found a problem if all of the files are covered by a keep_rules, and `-gsample N` does not match line count - there is no error message - I propose it mentions it cannot reach the sampling size.

guillaumekln: Thank. I added a warning in that case.

PR #411, opened 2017-10-23
update vocabularies when the new train-set's vocabularies are different

bwang-systran:
The new vocabulary is known in advance for dynamic dataset. The vocabulary update well be activated using an option ( --update_vocab), the default value is true.

### DIFF
--- a/docs/training/retraining.md
+++ b/docs/training/retraining.md
@@ -63,4 +63,4 @@ Another use case it to use a base model and train it further with new training o
 
 ## Updating the vocabularies
 
-In the spirit of infinite training, we could have different vocabularies in dynamic dataset and the pre-trained model. Instead of re-initializing the whole network, the pre-trained states of the common words in the new/previous dictionaries can be kept with option `--update_vocab`. This option is enabled by default.
+It is possible that we restart the training with a new dataset such as dynamic dataset, we could have different vocabularies in dynamic dataset and the pre-trained model. Instead of re-initializing the whole network, the pre-trained states of the common words in the new/previous dictionaries can be kept with option `-update_vocab`. This option is disabled by default and the update of word features isn't supported for instant.
--- a/docs/training/retraining.md
+++ b/docs/training/retraining.md
@@ -60,3 +60,7 @@ Additionally, the `-continue` flag retrieves from the previous training:
 ## Training from pre-trained parameters
 
 Another use case it to use a base model and train it further with new training options (in particular the optimization method and the learning rate). Using `-train_from` without `-continue` will start a new training with parameters initialized from a pre-trained model.
+
+## Updating the vocabularies
+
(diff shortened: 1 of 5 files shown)


bwang-systran:
If we start the trainning from a pre-trained model with a new train-set. It is possible that their dictionaries are not exactly the same, the difference could be found in the source side, in the target side or in both side.

With default activated option  `--update_vocab`, we are doing following updates to keep the common information (embedding, generator) from the pre-trained model.
For a word **_Word_a_** in dictionaries **_Dict_New_**:

1.  - Contains **_Word_a_** presents in **_Dict_Old_**: find **_Word_a_**'s index in **_Dict_Old_**, copy the all the pretrained information and save it in the new index in **_Dict_New_**.
2.  - Contains **_Word_a_** doesn't present in **_Dict_Old_**: initialise **_Word_a_**'s information and save it in the new index in **_Dict_New_**.

guillaumekln:
Thanks. Can you add your last comment as documentation in `docs/training/retraining.md` (you can add a new section "Updating the vocabularies").

onmt/train/Trainer.lua
```diff
@@ -77,6 +77,10 @@ local options = {
     [[When using translation-based validation metrics (e.g. BLEU, TER, etc.), also save the
     translation every this many epochs to the file `<save_model>_epochN_validation_translation.txt`.
     If = 0, will not save validation translation.]]
+  },
+  {
+    '-update_vocab', true,
```
guillaumekln: To not surprise existing users, this option should be disabled by default. Also update the documentation accordingly.

docs/training/retraining.md
```diff
@@ -60,3 +60,7 @@ Additionally, the `-continue` flag retrieves from the previous training:
 ## Training from pre-trained parameters
 
 Another use case it to use a base model and train it further with new training options (in particular the optimization method and the learning rate). Using `-train_from` without `-continue` will start a new training with parameters initialized from a pre-trained model.
+
+## Updating the vocabularies
+
+In the spirit of infinite training, we could have different vocabularies in dynamic dataset and the pre-trained model. Instead of re-initializing the whole network, the pre-trained states of the common words in the new/previous dictionaries can be kept with option `--update_vocab`. This option is enabled by default.
```
guillaumekln: `-update_vocab` instead of `--update_vocab`.

train.lua
```diff
@@ -131,6 +147,71 @@ local function loadModel(opt, dicts)
 
   checkpoint, opt, paramChanges = onmt.train.Saver.loadCheckpoint(opt)
 
+  if checkpoint.dicts ~= dicts and opt.update_vocab then
```
guillaumekln: `checkpoint.dicts ~= dicts` is always `true` as you are comparing pointers not content. I think it is reasonable to always execute this code when `opt.update_vocab == true` so just drop the first condition.

train.lua
```diff
@@ -131,6 +147,71 @@ local function loadModel(opt, dicts)
...
+    _G.logger:info(' * Found different vocabularies in the new train-set')
```
guillaumekln: Could this whole block code (under the `if opt.update_vocab then` condition) be in a separate function?

bwang-systran:
Some changes made as request.

PR #443, opened 2017-11-24
add support of features vocab updating

bwang-systran:
We update the generator of features when update_vocab is activated.


guillaumekln:
Could we also update the embeddings of the features as well?

guillaumekln:
Thanks!

PR #460, opened 2017-12-13
Integrate hooks with the rest translation server

panosk:


tools/rest_translation_server.lua
```diff
@@ -8,7 +8,7 @@
 ]]
 
 require('onmt.init')
-
+require('copas')
```
jsenellart: this does not belong to this PR, is this?
do you have an integration of copas in the rest server?

panosk: Hi @jsenellart ,

Actually, I just blindly included it after getting the error: 
`you must require copas before require'ing socket.http`
Without this require, the server won't start with hooks.

panosk: On a second glance, this requirement seems to occur because of the tree-tagger.lua hook which requires `httpclient`. Maybe we should just move the copas require to the tree-tagger.lua?

jsenellart: yes - it will make sense. Also, I will also independently look at integrating with copas in the servers.

PR #465, opened 2017-12-19
Fix detokenization when replaced target tokens contain spaces

guillaumekln:


guillaumekln:
@jsenellart FYI.

Detokenization has now the following interfaces:

1. `detokenize(opt, words, features)`
2. `detokenizeLine(opt, line)`

The first one is called whenever possible to avoid ambiguous cases when mixing placeholder replacement and case feature.

PR #467, opened 2017-12-20
Add optional custom features for domain control + small fix in server

panosk:
After trying this for a while (https://arxiv.org/pdf/1612.06140.pdf), the results are great and I think it would be a nice addition. The new `feats` field in the json request is completely optional and the user can add as many custom features (tags) as they like.


jsenellart:
Hi @panosk - great! I made a patch for integrating batches and it conflicts with your PR, could you please merge?

panosk:
Hi @jsenellart , done!

jsenellart:
thanks @panosk !

PR #475, opened 2017-12-28
B474

jsenellart:


vince62s:
@jsenellart what was wrong exactly with batch translation in the rest_server ? It was working fine for me (maybe with an older hash).
Also, if so, shouldn't we change the multi model rest server as well ?

jsenellart:
batch translation was simply not implemented - batch size option was misspelled and ignored. as a result - it was never batching

jsenellart:
yes - we have the same in `rest_multi_models.lua`. this could be observed if you were sending a very number of sentences to translate simultaneously - I will adapt too

PR #506, opened 2018-01-29
WIP introduce token batch size limit

vince62s:
The goal is to do this: http://forum.opennmt.net/t/memory-optimization/941

There is 2 limits: 
`-max_batch_size` = a number of sentences
`-max_tokens` = a number of tokens
When the first limit is reached the batch is full.
Example:
`-max_batch_size  160`
`-max_tokens  1800`
It will allow up to 160 sentences of short length up to 11 tokens
It will cap the number of sentences to a lower limit, ie 22 sentences of length 80

It works with normal training.
Should be fine with Dynamic Dataset also

@jsenellart @guillaumekln Need your help for the SampledDataset, I am not 100% sure.


jsenellart:
hello @vince62s, thanks for the initiative! To guarantee no OOM (and optimise full memory usage), it is a bit more complicated because the memory used by a sequence of size `sentence_length` is not exactly proportional to the product `batch_size*sentence_length` (which is the `token_count`) but also grow linearly with sentence length (because of some memory copy that are being done for some layers) - so we do have to reduce `batch_size` more. In other words, `max_mem=K1+(batch_size+K2)*sentence_length=K1+token_count+K2*sentence_length` => we should better select batch_size with `batch_size=(max_mem-K1)/sentence_length-K2`. I will try adding an estimator of `K1` and `K2` at the beginning of the process that will refine your approach.

vince62s:
I think I get it but it depends on the max_seq_len given by the user and the distribution of the data. So I am wondering if this is really worth it.
Empiracally I found a few working settings depending on the layer size, number of layers, embedding and sequence length. But then there will be the issue of images, audio, ....
While talking about this, there is nice improvement to save memory which is embedding sahring between src and tgt when using shared vocab. Second one is Adafactor optimizer in place of Adam to save a lot of memory.

jsenellart:
I have something working, I am launching some validation runs (for memory, for quality, I am sure you will do it better) and will push my code.

jsenellart:
the idea of predicting memory is more complicated than I thought - essentially because the `cutorch.freememory` is a poor indicator on actual available memory (especially with `TH_CACHING_ALLOCATOR=1` but more generally because of memory fragmentation). What I did instead is to initialize all of the internal memory structures with maximal size (max batch, and max tokens). In this phase most of the memory is allocated, I have still few variations that are coming I think from attention layer - but if the process has to fail from oom, it is likely it fails early in the preallocation phase.

vince62s:
Jean, there is still the SampledDataset to modify.
I was a bit shy to do this one, seems a bit tricky

PR #539, opened 2018-05-04
Return source and target tokens with attention table

panosk:


jsenellart:
Hi @panosk, thanks for the PR and sorry for late follow-up. Can you just update the `CHANGELOG.md`?

jsenellart:
Thanks!

PR #542, opened 2018-05-14
[WIP] convert to short and introduce padding in the tensor storage

jsenellart:
option to quantize weights in model into INT16 (short) - reduces t7 size by 2.

### DIFF
--- a/tools/release_model.lua
+++ b/tools/release_model.lua
@@ -41,6 +41,7 @@ local function isModel(object)
   return torch.type(object) == 'table' and object.modules
 end
 
+local sharedStorage={}
 local function releaseModule(object, tensorCache)
   tensorCache = tensorCache or {}
   if object.release then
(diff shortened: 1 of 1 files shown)


tools/release_model.lua
```diff
@@ -39,7 +46,15 @@ local function releaseModule(object, tensorCache)
   if object.release then
     object:release()
   end
-  object:float(tensorCache)
+  object:apply(function (m)
+    for k, v in pairs(m) do
+      if torch.isTensor(v) and v:dim()>0 then
+        v:mul(1000)
+        m[k] = v:short()
```
guillaumekln: Note that this cancels weights sharing. This concerns for example word embeddings in bidirectional encoders.

See https://github.com/OpenNMT/OpenNMT/commit/de97aa259c01f15769a4550637bbd807b852d490.

tools/release_model.lua
```diff
@@ -39,7 +46,15 @@ local function releaseModule(object, tensorCache)
   if object.release then
     object:release()
   end
-  object:float(tensorCache)
+  object:apply(function (m)
+    for k, v in pairs(m) do
```
guillaumekln: There is already the same loop a few lines below. The body could be moved there.

PR #545, opened 2018-05-23
Fully ignore placeholders when learning BPE

guillaumekln:
Until now, each placeholder had its own rule in the BPE model for the encoding to not split such token. However, all encoding scripts (here and in the Tokenizer) simply skip the protected sequence, making these rules unnecessary.

For consistency with our own tools, I propose to ignore all placeholders when learning the BPE model.

@jsenellart Does it make sense? Or would it be better to configure the behavior regarding placeholders with an option?


jsenellart:
yes it makes sense.

